PR #789, opened 2013-05-20
Dox warning feature

honestbleeps:
This pull request consists of a few very simple changes.  One known issue is the copy: I imagine you want to edit/change this copy, maybe have a lawyer look at it, etc - even if it's just a sentence or two.  I've worked with plenty of large entities in the past and all text must be vetted, so I didn't spend much time on the copy.  I imagine it might ultimately end up in a resource file as well (for internationalization), but I wasn't sure.

That said, the functionality is simple: a debounced function will scan comments as they're being typed and look for strings that appear like they *might* be personal information.  Currently it scans for phone numbers, email addresses, facebook and twitter profile links.

I'm aware the rate of false positives may be high, but the annoyance level is relatively low here as it's unobtrusive.

I did investigate using a hover-bubble for the rules link, but found it may make more sense to simply link to the rules page. Here's a simple screen shot example:

screen shot sample

I had originally envisioned something prettier looking, but it got in the way enough that it felt obtrusive, so I opted for this route instead.

A few notes:

- Regex addition is simple, there's a dictionary of them that you can simply add to, the script will automatically check them. If you want to add tumblr links, etc - it's trivially easy.

- I'm storing the types of dox-looking stuff found in an array in the javascript, which could easily be displayed to the user, but felt it wasn't worth the screen real-estate now that I was no longer doing any sort of a hover bubble or a more obtrusive setup.

- I'm certainly open to changing the approach, this is just a quick stab!


kjoconnor:
Looks like we let this one go too long - I'm going to close it for now. If anyone's interested in getting this merge-able again please reopen.

PR #790, opened 2013-05-20
added random placeholder code to demonstrate the idea

honestbleeps:
This is based off of a really simple idea I was discussing briefly with chromakode.

The copy is absolutely silly and up for change, of course... it's late and I came up with each one in a few seconds...

The concept is simple: At random, put in a placeholder message that jars the user just enough to get them to think twice about what they're about to post.  Maybe it's something funny or something positive - maybe it's a flat out reminder that civil discourse is what is supposed to separate reddit from other communities.

Why random?  Because it's easier to learn to tune out if it's there every time.

From a tech standpoint, this would likely need to be updated somehow to handle internationaliization, but I at least wanted to get the idea in front of you for review/discussion.

### DIFF
--- a/r2/r2/public/static/js/reddit.js
+++ b/r2/r2/public/static/js/reddit.js
@@ -859,6 +859,7 @@ function comment_reply_for_elem(elem) {
         form.attr("id", "commentreply_" + thing_id);
         form.find(".error").hide();
     }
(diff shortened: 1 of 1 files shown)


kjoconnor:
I think we won't be using this at the moment, thanks though!

PR #865, opened 2013-07-29
Update classes on thing when selecting linkflair

rasher:
To give a proper instant preview, the linkflair classes need to be added to the thing div when flair is selected. As well, any existing classes need to be removed first (especially important when removing flair).


kjoconnor:
Looks like we let this go too long and now has drifted pretty far from `master`. If anyone's interested in making the necessary changes to make it merge-able again, we can reevaluate.

Thanks!

PR #877, opened 2013-08-09
Remove forced margin between wiki and sidebar

rolmos:
Currently, the wiki content is given a margin with the sidebar using a right margin of 325px. This value will always depend on the sidebar's width.

Since the rest of Reddit's content flows with the sidebar this element should use the same behaviour. This way, the wiki maintains a margin, but doesn't use that huge margin value.

To test the benefits: delete the sidebar node in a current default wiki page. The wiki will maintain that huge space on the right unnecessarily.


kjoconnor:
Looks like we let this go too long - if you're still interested in having this merge please rebase off master and reopen. Thanks!

PR #889, opened 2013-08-26
always allow "unvoting" on things, regardless of age

jcs:
Preventing upvotes or downvotes for old items makes sense, but users should be allowed to remove votes they've previously cast if they're determined to (which would require directly submitting to the API, no UI changes here).


kjoconnor:
We don't allow any type of voting action on old items, both for reasons of we don't want to have old items be acted on but also to leave ourselves room to move old items to cold storage in the future.

PR #890, opened 2013-08-27
Added help text for stickying posts in linkcommentssettings.html

AuroraAri:
This change adds a help text visible to moderators for "sticky this post" in the same way that contest mode has. It does not appear for users who don't have the ability to sticky the post in the first place, of course. :) All for the sake of congruency and on-page documentation!

sticky-help


bsimpson63:
Sorry it took so long to get to this. The equivalent tooltip for contest mode was removed in  a5e8661fba199198f1aca0cc8356cd149e0549dd, so adding a tooltip for sticky-ing doesn't make as much sense anymore.

PR #910, opened 2013-10-16
Remove explicit height on .subtoolbar and fix background on selected menu item.

Jacse:
With a fixed height on .subtoolbar menu items disappear on iPods, iPhones etc. (all devices narrower than 444px).
Browsers display it jsut fine without the height set.

I also changed a padding by one pixel, which makes the circular background fit inside the container and not be cut off.

pullrequest


FlashYoshi:
+1 seems to do the job, shouldn't this be merged? Or just closed as it might be irrelevant now?

kjoconnor:
Yeah, we've deprecated i.reddit.com in favor of m.reddit.com so I'll close this. Thanks though!

PR #942, opened 2014-01-04
Prevent animation queuing of search box.

Rhym:
Added jQuery .stop() method to the slide animations for '#searchexpando'.


kjoconnor:
Is this still happening? Where is this an issue? Please reopen if you're interested in continuing with this.

Rhym:
This is still relevant going on a year and a half later :P

You can replicate this by focusing in and out of the search bar in rapid succession. Well pretty much anything that uses the slideUp/Down() jQuery methods. They should be prefixed with a stop(false, true) to prevent animation queuing.

PR #970, opened 2014-02-07
Links within Reddit are manipulated to end in .compact if the

mmulqueen:
Links within Reddit are manipulated to end in .compact if the user is browsing in compact mode.

Fixes: https://github.com/reddit/reddit/issues/129


kjoconnor:
We've deprecated compact views in favor of m.reddit.com. Thanks though!

PR #987, opened 2014-03-09
Added banned page class for banned users

jaffamol:
Very minor change to include a banned user class to page_classes as requested below.

http://www.reddit.com/r/ideasfortheadmins/comments/1zz2u1/allow_customisation_of_css_for_banned_users


kjoconnor:
I think we have something in the works internally to take care of this, and since this is out of sync with master we can go with our fix hopefully in the near future. Thanks though!

PR #1003, opened 2014-04-12
Add /r/ to subreddit name in more places.

listen2:


kjoconnor:
I don't think we need to make this change. Thanks though!

PR #1018, opened 2014-04-22
Added /r/ to other HTML templates

austinoneil:
I noticed that on some of the pages, /r/ is not added to the subreddit.  I can't say that every page has /r/, but it's a good start, right?

Ignore the redirect for mobile users commit.  I made that on an earlier day and I don't know how to get rid of it.  Here's how to ignore that commit: http://mispdev.blogspot.com/2013/02/github-cherry-picking-commits-from-pull.html  If you know how to cherry pick commits to be used in a pull request, please let me know.


bsimpson63:
Sorry it took so long to comment on this. Making changes like this is more of a design/product decision, so I'm going to close this pull request. Thanks for the contribution.

PR #1035, opened 2014-05-07
Change dropdown cursor to pointer

DoctorMcKay:
Change the cursor when hovering over a dropdown activator to a pointer (hand) for consistency with other clickable elements.


bsimpson63:
This has been open for a while and I don't think there's much interest in making this change. Sorry.

PR #1037, opened 2014-05-08
added a more specific selector on hiding / showing comment

honestbleeps:
the existing selector is inadvertently too greedy. Specifically causing this issue:

http://www.reddit.com/r/RESissues/comments/22b8ob/bug_after_collapsing_then_reopening_a_comment/


bsimpson63:
It's been a while, but is this still an issue?

honestbleeps:
I think we released a fix to RES - @andytuba may remember for sure - but in actuality the selector was too general no matter what, in the sense that it could cause you guys problems down the line too.  No idea how your codebase has since evolved, though.

bsimpson63:
Is it ok if I close this?

PR #1054, opened 2014-05-30
Add header bar first link left padding

DaAwesomeP:
The first link in the top bar is nudged right against the border of my monitor. On my laptop, it appears cut off a little (the plastic frame overlaps the LCD). It also just looks better.

Before:
reddit before
After:
reddit after

I added a `padding-left` of 4px. I used the `first-child` pseudo-element selector to ensure future proofing (any link there will also have that padding).


kjoconnor:
I don't think we need to make this change - thanks though!

PR #1058, opened 2014-06-04
Updated cssfilter.py

brettwilcox:
Added the css filter property per mozilla guideline - https://developer.mozilla.org/en-US/docs/Web/CSS/filter

Excluded the URL in the spec for safe functions since we do not want users to have the ability to use images outside of reddit uploaded images.


spladug:
:eyeglasses: @spladug

(I'm [taking this on for review and merging](https://github.com/reddit/reddit/wiki/The-Code-Salon))

spladug:
Hi there, thanks for your patch! Unfortunately, during security review we determined that we can't enable this property at this time because it would allow XSS on IE versions 9 and below.  We currently allow IE9 to see subreddit styles and our current traffic analysis shows we're not ready to break that aspect of the site for those users yet.  We'll re-evaluate this in a few months to see if the usage of IE9 has dropped sufficiently.

Here's the full explanation from the astute @JordanMilne:

> We allow IE9 and above to use subreddit styles, but in IE versions below 10 `filter` refers to DirectX filters. At least one filter will cause cross-origin requests ([`AlphaImageLoader`](https://msdn.microsoft.com/en-us/library/ms532969%28v=vs.85%29.aspx)), and since the syntax for filters came in IE5.5, way before "real" css functions, IE is pretty lax about how it parses that property. For ex:

```css
*{
    filter:progid:DXImageTransform.Microsoft.AlphaImageLoader (src='http://saynotolinux.com/tablecat.png', sizingMethod='scale');
}
```

> will get by our css filter, and IE9 will still accept it and make the request regardless of the space between the filter name and the opening bracket.

M108:
> We'll re-evaluate this in a few months to see if the usage of IE9 has dropped sufficiently.

What's the status on this now?

spladug:
too high

13steinj:
http://niceme.me

When it comes to that though; why not just disallow filters that make cross origin requests as the value (Unless I'm misunderstanding the capabilities of the current filter)

spladug:
> When it comes to that though; why not just disallow filters that make cross origin requests as the value (Unless I'm misunderstanding the capabilities of the current filter)

See [above](https://github.com/reddit/reddit/pull/1058#issuecomment-76466180) where I quoted Jordan's explanation of a bit of CSS that would get by our filter (because it's taking advantage of older IE's loose and incorrect interpretations of CSS syntax) and still be valid enough in IE to generate a cross-origin request.

1j01:
Can't we just disallow `/progid\s*:/i` and/or `/DXImageTransform/i` for the `filter` property?

walle303:
> too high

It has been over a year since you re-evaluated, has IE9 usage dropped sufficiently by now?

esr360:
Hello, please can you re-consider implementing the filter property? It's not worth preventing its use just for IE9 users.

PR #1063, opened 2014-06-06
Add the ability to shortlink comments

empyrical:
Adds the ability to shortlink comments, like: redd.it/c/c12345

And if a user is viewing an individual comment, this will show a box under the submission shortlink showing the comment shortlink.


Dakta:
This never get merged? I think it's a great feature.

empyrical:
Yeah, closed it because it was unmergable due to the commits that have happened since then and don't have the time right now to fix it up. If the admins are interested it could be re-opened though

PR #1084, opened 2014-07-03
Allow sorting userpage overview, submissions, and comments by subreddit

archaicwisdom:
This should allow users with gold to sort their userpage by subreddit in order to view comments, submissions, or both, from just a specific subreddit. This is similar to sorting saved links by subreddit.


bsimpson63:
Sorry this has sat for so long without comment, but I don't think this is a feature we want to add at this time. This would cause a lot of issues for how we do caching and precomputed queries, so if we did decide to add it we would need to think about performance implications.

Thanks for your contribution.

PR #1085, opened 2014-07-08
Remove a back button-breaking hack in .compact

Sidnicious:
Scrolling the document down by a pixel used to hide the navigation bar in iOS Safari. It doesn't do anything in newer versions of iOS Safari, but it does make iOS Chrome lose its scroll position when you go back to a .compact page.

http://www.reddit.com/r/bugs/comments/2a5ee0/when_using_compact_on_ios_chrome_if_i_go_back_the/

### DIFF
--- a/r2/r2/public/static/js/compact.js
+++ b/r2/r2/public/static/js/compact.js
@@ -41,10 +41,6 @@
 })(jQuery);
 
 $(function() {
-    if ($(window).scrollTop() == 0) {
-        $(window).scrollTop(1);
-    }
-    ;
     /* Top menu dropdown*/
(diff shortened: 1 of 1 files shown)


JordanMilne:
@dwick @ajacksified Any UX reason why we might still want to do this? Verified that this code currently does nothing in Safari on iOS 8, but always scrolls to the top on clicking back in iOS chrome.

bsimpson63:
Sorry this has sat for so long without comment. The .compact interface isn't getting much development attention these days, so I'm going to close this pull request. Sorry, and thanks for the contribution.

Sidnicious:
No worries, thanks for the response :+1: !

PR #1093, opened 2014-07-18
Fix flairselection dropping under elements

creesch:
Changing the opacity of elements messes with their z-index. I got the following screenshot from a person using our mod toolbox extension:

- http://i.imgur.com/DqHffaX.png 

On investigation I found that it was not toolbox or RES related but rather the fact that shadowbanned submissions have an opacity of 0.7. Removing the opacity and setting a background resulting in the same color instead fixes this issue.


bsimpson63:
This has been open for a while without activity so I'm going to close it. If this is still an issue and you'd like us to revisit it just leave a comment here.

PR #1108, opened 2014-08-10
Fixed widget creation page

joshuaferrara:
Originally, whenever a user tried to configure a widget on the configuration page, the document's body would be set to an HTTP error code. This fixes this issue.


JordanMilne:
Whoops, just noticed this. Something almost identical went out in https://github.com/reddit/reddit/commit/0e276784164fd4f5a68f455316e9faf7f21c9401. Thanks for the PR!

PR #1113, opened 2014-08-20
Remove 7em padding-right on advertising links

Andygmb:
Causes the text in the title of the advertising to pile up on smaller resolutions, pushing content down & being harder to read.

Before:

------------------

After:

### DIFF
--- a/r2/r2/public/static/css/reddit.less
+++ b/r2/r2/public/static/css/reddit.less
@@ -1465,7 +1465,6 @@ body.with-listing-chooser.explore-page #header .pagename {
 .organic-listing .link,
 .organic-listing .link.compressed,
 .organic-listing .link.promotedlink {
-    padding-right: 7em;
     padding-left: 2px;
     margin-bottom: 0px;  
 }


13steinj:
@bsimpson63 Are you sure it's only on the fron page? If so, you can add

    .front-page {
        .organic-listing .link,
        .organic-listing .link.compressed,
        .organic-listing .link.promotedlink {
            padding-right: 55px;
        }
    }

after @Andygmb 's change. I used px instead of em because em is a relative unit; and changes based upon what a user can set in their browser.

If you  find the issue occurring on other pages, you could could add them to the first line separated by a comma.

bsimpson63:
I'm not sure that it's always on the frontpage, and it might change in the future. It'd probably be most straightforward to add a class to the organic listing/spotlight box when the next/prev arrows are present, and then add the padding if that class was present.

bsimpson63:
Closing due to inactivity. If you'd like us to revisit this just leave a comment.

PR #1122, opened 2014-09-06
Replaced SQL functions' params with named params

Decker108:
Changed SQL functions to use named parameters instead of positional parameters, which improves readability. Named params are implemented in all currently supported versions of Postgresql.


kjoconnor:
We generally don't make changes like this just for readability's sake, so I'm going to close this. Thanks for the submission though!

PR #1138, opened 2014-09-29
Extend Quota for registered users

kuletxcore:


bsimpson63:
I'm not sure this is something we want to do, and it's been inactive for quite a while so I'm going to close. Leave a comment if you'd like us to revisit this idea.

PR #1192, opened 2014-12-07
Add votable to link api response

kennydude:
Useful for apps to determine if item can be commented/voted on

Only thing I'm not too sure on is the name, as something like `archived` might be better


kemitche:
`archived` was added separately.

PR #1199, opened 2014-12-11
Added word-wrap property

zedadex:
This property is a useful one for positioning. I find myself needing it in cases where text-wrap does not do what I need, and would appreciate it being added to the whitelist. Thanks for your consideration.


bsimpson63:
Sorry, but I'm going to close this for inactivity. If you'd like us to consider it just leave a comment.

PR #1202, opened 2014-12-13
Feature/markdown preview

kochis:
fixes #133
closed #1195 in favor of a back-end approach.

More detailed description and screenshots here: http://www.reddit.com/r/ideasfortheadmins/comments/2oaw31/markdown_preview_for_comments_and_self_posts/

Using a tab-base UI approach similar to the github preview would probably be preferable. 

(this PR is a re-issue of #1201

### DIFF
--- a/r2/r2/controllers/api.py
+++ b/r2/r2/controllers/api.py
@@ -1675,15 +1675,8 @@ class ApiController(RedditController):
         text=VMarkdown('text'),
     )
     @api_doc(api_section.links_and_comments)
-    def POST_preview_comment(self, commentform, jquery, text):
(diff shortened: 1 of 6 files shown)


JordanMilne:
Just had a look at this, if you want to reopen #1195 and have it use https://www.npmjs.com/package/reddit-text-js, that'd be preferable to a back-end approach. We shied away from using SnuOwnd before because some XSS issues were still unfixed upstream, but we've been using a fork of it on m.reddit.com with a sanitizer in front for a while now. 

From my testing it now behaves nearly identically to the backend version, so hitting the backend is unnecessary.

PR #1203, opened 2014-12-15
Change SubredditSelector to use better wording

allthefoxes:
Currently, the subreddit selector when submitting a new link "/r/subreddit/submit" says "Popular Choices" and then lists the subreddits you are subscribed to.

I find this to be a big misnomer, as they are not popular choices at all, they are just a list of your subscribed subreddits. "Popular choices" implies that these are things other people have selected as well.

I know, it's super nitpicky, but I thought it might make more sense this way.


bsimpson63:
Merged as 8fdc79956401157953a7dbe010cd999d7ca90c44

PR #1211, opened 2014-12-17
check 'limit my search to X' by default

samertm:
I probably check this box 90% of the time I do a search on a subreddit
page. The last time I checked (heh), RES checked this box by default
as well. Thoughts?

### DIFF
--- a/r2/r2/public/static/js/reddit.js
+++ b/r2/r2/public/static/js/reddit.js
@@ -1215,12 +1215,12 @@ $(function() {
             $("#searchexpando").slideDown();
         });
 
-        // Default to true for the 'restrict search' checkbox.
-        $('#searchexpando input[name="restrict_sr"').
-            prop("checked", !(store.get('searchexpando.input.checked') === false));
-        $('#searchexpando input[name="restrict_sr"').change(function() {
-            store.set('searchexpando.input.checked', this.checked)
-        });
+        // Store the user's choice for restrict_sr
+        $('#search input[name="restrict_sr"]')
+          .change(function() {
+            store.set('search.restricted_sr.checked', this.checked)
+          })
+          .prop("checked", !!store.get('search.restricted_sr.checked'));
 
         $("#search_showmore").click(function(event) {
             $("#search_showmore").parent().hide();
--- a/r2/r2/public/static/js/reddit.js
+++ b/r2/r2/public/static/js/reddit.js
@@ -1215,6 +1215,13 @@ $(function() {
             $("#searchexpando").slideDown();
         });
 
+        // Default to true for the 'restrict search' checkbox.
+        $('#searchexpando input[name="restrict_sr"').
+            prop("checked", !(store.get('searchexpando.input.checked') === false));
+        $('#searchexpando input[name="restrict_sr"').change(function() {
+            store.set('searchexpando.input.checked', this.checked)
+        });
+
         $("#search_showmore").click(function(event) {
             $("#search_showmore").parent().hide();
             $("#moresearchinfo").slideDown();
--- a/r2/r2/templates/searchform.html
+++ b/r2/r2/templates/searchform.html
@@ -66,7 +66,7 @@
   %elif thing.simple:
   <div id="searchexpando" class="infobar">
       % if not isinstance(c.site, DefaultSR):
-      <label><input type="checkbox" name="restrict_sr" tabindex="21">${_('limit my search to %(path)s') % dict(path=c.site.path.rstrip('/'))}</label>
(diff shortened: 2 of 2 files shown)


umbrae:
I've pulled some data on this, and it looks like it is the case that this box is checked more often than not if within a specific subreddit - around 68% of the time.

I'd recommend we merge this. Any qualms?

:eyeglasses: @xiongchiamiov @Deimos

umbrae:
> It also somewhat depends on whether we want to drive discovery of content within subreddits, or between them.

That question is a good one. If/when we redesign search results, I think we could take that into account. For now though, this is only active when someone is actually on a subreddit, so homepage searches will still have the discoverability people are likely expecting. 68% checking a non-default box is pretty significant to my eye.

Deimos:
No objections here, I definitely personally check it a lot more often than not.

xiongchiamiov:
From the data you pulled, I'm definitely in favor of changing it.

There are, of course, going to be some people who will be annoyed at this.  Adding another preference seems like a bad idea, given our tendency to become `about:config`.  I think a reasonable idea is to make the box "sticky" - that is, store its checked or unchecked state in a cookie, so it stays how you last set it.  But my inclination is to discuss and potentially ship that separately from this.

umbrae:
If we did do that, I'd prefer storing it in localStorage so it's not passed along in every HTTP request. That would be pretty nice but would be a bit more work than this PR.

@samertm - what's your interest level in taking a crack at that? Any? If so, that'd be awesome, but I also recognize this is just for fun for you and not your job.

samertm:
Sure, I can work on that this weekend.

spladug:
For reference: we have a localstorage wrapper in the JS code already: https://github.com/reddit/reddit/blob/master/r2/r2/public/static/js/lib/store.js

And an example of it in use: https://github.com/reddit/reddit/blob/master/r2/r2/public/static/js/adminbar.js#L97-L101

umbrae:
:sparkles: :tada: :cake: :tada: :sparkles: 

https://www.reddit.com/r/changelog/comments/2tb6gd/reddit_change_unread_counts_on_your_inbox/

https://github.com/reddit/reddit-private/commit/46807f4c45fece0f64bc2b6101a740480a5a2edc

spladug:
:running: to clean up my code review dashboard

umbrae:
@samertm Are you interested in tackling this or should I close this out and add a ticket for us internally?

samertm:
Hey, sorry for taking ages on this! I haven't set up a dev environment for reddit to test this, but it works against the live site (substituting a localStorage shim for store).

umbrae:
Hey @samertm,

This works for me! I made some slight adjustments, want to take a quick look at the diff here https://gist.github.com/umbrae/0546cb66499b63836960

Adjustments:

* There was a bug in your selector - needed a trailing `]`
* Now that this is sticky (which I think is a great solution), I think we may actually want to default to having it not checked. The reason being that this can encourage serendipitous discovery of other subreddits, which is handy for new users. If a user checks it, then the intent is pretty clear that they intend to search within a community, so serendipitous discovery is less of a concern.
* Some small style/cleanup stuff.

Thoughts on that?

samertm:
That sounds good to me. I installed your patch and reverted the commit that sets "checked" on the input tag. Sorry for taking ages for such a small change :o)

umbrae:
@samertm I've merged this change. Thanks for the inspiration. Would you like an open sourcerer badge on reddit? If so, let me know your username.

umbrae:
(Link: https://github.com/reddit/reddit/commit/42ca7212e5549318f7173250de66936ee352f2e6)

samertm:
Thanks for installing this change! The search restriction not being sticky has bugged me for ages. I think this change was too small to merit an "open sourcerer" badge, but I'll see if there's anything else I can sink my teeth into to really earn it :)

xiongchiamiov:
It's something that's been bothering many people for a long time; in terms of impact, I don't think it's small at all. :)

BTW, [here](https://www.reddit.com/r/changelog/comments/2vfy5b/reddit_change_limit_my_search_to_rsubreddit_now/) is the changelog post if you want to keep up on feedback.

andytuba:
Small fallout bug on the /search page: 

If you visit a search page like `.../search?restrict_sr=on` (e.g. filter subreddit's posts by flair) but haven't personally changed the option, then this snippet will unexpectedly turn off restrict_sr.

Should this be resolved by only selecting the sidebar search, or by testing if the checkbox was already checked server-side before touching it?

xiongchiamiov:
Hmm, it kinda makes sense to have the restriction apply no matter where it's done; that way, you'll get users who search, *then* realize they wanted to restrict it.  OTOH, having that sticky over to a separate search box is kinda weird.

I'm inclined to target only the sidebar search.

PR #1218, opened 2014-12-21
Removed white space in fetch_more() function

willempienaar:
Just removed two blank lines of white space.


bsimpson63:
Thanks for the pull request, but we generally don't accept whitespace only changes. We try to clean up whitespace issues when modifying the nearby code.

PR #1219, opened 2014-12-22
Added missing closing head tag

Aedx:


xiongchiamiov:
Thanks: bca1c9e7a4207525c8dbd8d5939afede66a36b77 .  We've got a really long cache time on that, so you probably won't see the effect until the caches restart or we invalidate them or whatever.

If you let me know your reddit username, I'll give you the contributor trophy.

PR #1222, opened 2014-12-26
Adjusted maxlength

wnklmnn:
Username can not be over 20 char long.


bsimpson63:
Closing for inactivity. If you'd like to reopen it just leave a comment here.

francis-liberty:
I've taken a look, xiongchiamiov was right, but login.html and loginformwide.html were already corrected. I've looked through templates, only wnklmnn's wikipagesettings.html and flairlist.html still uses 32.

PR #1226, opened 2015-01-03
append .compact to reddit.com links in mobile version

Mtillmann:
Adds .compact to all reddit.com links in mobile version to avoid visiting the desktop site when browsing meta subreddits and following links to other threads etc


bsimpson63:
Thanks for the pull request, but we're not going to accept this. .compact mode is not getting much active development these days. I think we'd maybe want to take a more targeted approach to rewriting internal urls.

Sorry it took so long to reply to this.

PR #1228, opened 2015-01-07
Updated copyright headers for 2015 - Happy New Year!

LiterallyElvis:
I went file by file and updated every copyright header that mentioned reddit as being the copyright holder. This does not apply to many of the things in /r2/r2/public/static/js/lib since other parties are listed as the copyright holders there.

Also apparently [Atom](http://atom.io) likes to strip whitespace, so there are a few lines like that.


spladug:
Thanks for the patch. We generally try to avoid making multiple types of changes in the same commit so that the blame can be more informative.  

af09fa8dee69bef4f65a1662d9ad91c2329946e1

PR #1232, opened 2015-01-08
Update api.py with new date 

alchemaLabs:
such 2015


spladug:
Thanks for the patch. We usually like to do this as a single huge batch to avoid cluttering up the history.

af09fa8dee69bef4f65a1662d9ad91c2329946e1

PR #1234, opened 2015-01-09
Add scheme to URL input if not present on losing focus

Exupery:
This is a fix for #1214 - which is caused by browsers that validate URL input fields preventing submission if no scheme is present, e.g. `example.com` instead of `http://example.com`. When an input field loses focus this will check if a scheme is present and prepend the input with `http://` if not.

### DIFF
--- a/r2/r2/public/static/js/reddit.js
+++ b/r2/r2/public/static/js/reddit.js
@@ -1229,6 +1229,17 @@ $(function() {
             $("#moresearchinfo").slideUp();
             event.preventDefault();
         });
+
+        /* Add a scheme to URL input fields if not already present */
+        $("input[type=url]").blur(function() {
+            var input = $(this).val();
+            if (input && input.length > 0) {
+                var schemePresent = input.match(/^[a-z]+:\/\/.*/i);
+                if (!schemePresent) {
+                    $(this).val("http://" + input);
+                }
+            }
+        });
         
         /* Select shortlink text on click */
         $("#shortlink-text").click(function() {
--- a/r2/r2/public/static/js/reddit.js
+++ b/r2/r2/public/static/js/reddit.js
@@ -1229,17 +1229,6 @@ $(function() {
             $("#moresearchinfo").slideUp();
             event.preventDefault();
         });
-
-        /* Add a scheme to URL input fields if not already present */
-        $("input[type=url]").blur(function() {
-            var input = $(this).val();
-            if (input && input.length > 0) {
-                var schemePresent = input.match(/^[a-z]+:\/\/.*/i);
-                if (!schemePresent) {
-                    $(this).val("http://" + input);
-                }
-            }
-        });
         
         /* Select shortlink text on click */
         $("#shortlink-text").click(function() {
--- a/r2/r2/controllers/api.py
+++ b/r2/r2/controllers/api.py
@@ -469,6 +469,7 @@ class ApiController(RedditController):
 
         if kind == 'link':
             if not url or form.has_errors("url", errors.NO_URL, errors.BAD_URL):
+                form.set_error(errors.BAD_URL, 'url')
                 return
 
             if form.has_errors("url", errors.DOMAIN_BANNED):
--- a/r2/r2/templates/newlink.compact
+++ b/r2/r2/templates/newlink.compact
@@ -67,7 +67,7 @@ ${thing.formtabs_menu}
 <div class="spacer">
   <%utils:round_field title="${_('url')}" id="url-field">
     <input name="kind" value="link" type="hidden"/>
-    <input id="url" name="url" type="url" value="${thing.url}"/>
+    <input id="url" name="url" type="text" value="${thing.url}"/>
     ${error_field("NO_URL", "url", "div")}
     ${error_field("BAD_URL", "url", "div")}
     ${error_field("ALREADY_SUB", "url", "div")}
--- a/r2/r2/templates/newlink.html
+++ b/r2/r2/templates/newlink.html
@@ -71,7 +71,7 @@ ${thing.formtabs_menu}
 <div class="spacer">
   <%utils:round_field title="${_('url')}" id="url-field">
     <input name="kind" value="link" type="hidden"/>
-    <input id="url" name="url" type="url" value="${thing.url}" required>
+    <input id="url" name="url" type="text" value="${thing.url}" required>
     ${error_field("NO_URL", "url", "div")}
     ${error_field("BAD_URL", "url", "div")}
     ${error_field("DOMAIN_BANNED", "url", "div")}


r2/r2/public/static/js/reddit.js
```diff
@@ -1229,6 +1229,17 @@ $(function() {
             $("#moresearchinfo").slideUp();
             event.preventDefault();
         });
+
+        /* Add a scheme to URL input fields if not already present */
+        $("input[type=url]").blur(function() {
```
xiongchiamiov: Hmm, I don't know if we want to do this everywhere - it seems like a reasonable thing to do, but that sort of wide-reaching change also has a tendency to unexpectedly break things.  It might be best to start by targeting the submission page specifically, then widening it as we find more areas it'd be useful.

r2/r2/public/static/js/reddit.js
```diff
@@ -1229,6 +1229,17 @@ $(function() {
...
+            var input = $(this).val();
+            if (input && input.length > 0) {
+                var schemePresent = input.match(/^[a-z]+:\/\/.*/i);
+                if (!schemePresent) {
+                    $(this).val("http://" + input);
```
xiongchiamiov: There's a part of me that wants to default to https, even if there are a bunch of sites that don't support it. But that's probably a bad idea. :)

r2/r2/public/static/js/reddit.js
```diff
@@ -1229,6 +1229,17 @@ $(function() {
             $("#moresearchinfo").slideUp();
             event.preventDefault();
         });
+
+        /* Add a scheme to URL input fields if not already present */
+        $("input[type=url]").blur(function() {
+            var input = $(this).val();
+            if (input && input.length > 0) {
+                var schemePresent = input.match(/^[a-z]+:\/\/.*/i);
```
xiongchiamiov: Using [the RegExp constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp) would get rid of those nasty backslash escapes:

    RegExp('^[a-z]+://.*', 'i')

xiongchiamiov:
Just adding the scheme directly is a little bit visually jarring.  I'm not sure what we'd want to do about that, though; I have an idea in my head of a form that has a greyed-out default protocol that gets overwritten when your link includes a protocol of its own, but I can't find an example of that.

:eyeglasses: @boriss @ajacksified for some more qualified feedback.

Boriss:
@xiongchiamiov Ideally, users submitting links wouldn't need to deal with missing schemes - submitting "cnn.com" would simply work with valid schemes added *after* - not before or during - link submission.  The submitted link would go to "http://www.cnn.com/" - much like entering cnn.com in the URL bar of a browser would.  If we're prepending the input with http:// *after* submission, we're meeting the link-submitter's expectations ("my submitted link worked without problems") and the link-clicker's expectations ("this link directs me properly").

r2/r2/public/static/js/reddit.js
Exupery: Initially I thought the same, but after checking it looked like this and the corresponding one in [newlink.compact](https://github.com/reddit/reddit/blob/77133cf7ee424267e608ee904d3df5062a435294/r2/r2/templates/newlink.compact#L70) are the only input boxes that uses the `url` type presently: `grep -rn 'type="url"' r2/ --exclude \*css` i.e. there shouldn't be anything for it to potentially break - it would impact future url inputs added (though presumably if it made sense here it would make sense in future ones)

Exupery:
@Boriss there's already appropriate code in place to do exactly the behavior you describe, if you repeat the [steps to reproduce](https://github.com/reddit/reddit/issues/1214#issue-52522611) with a browser that *doesn't* attempt to validate URL inputs (i.e. IE9) you can enter just `cnn.com` and submit it without issue.

The problem comes in to play because most modern browsers will only allow a valid URL if an input is marked as the `url` type. This can easily be avoided by changing this to the more generic `text` type and relying on the existing server-side code to validate (including adding a scheme if needed). i.e. change [newlink.html#L74](https://github.com/reddit/reddit/blob/77133cf7ee424267e608ee904d3df5062a435294/r2/r2/templates/newlink.html#L74) and submit `cnn.com` and it will submit a new link as `http://www.cnn.com/` - though this has some potential mobile-unfriendliness mentioned in my initial [comment](https://github.com/reddit/reddit/issues/1214#issuecomment-68536758).

Boriss:
@Exupery Well, if the only negative is that mobile users may lack a ".com" key - but the positive is that desktop users don't get the annoying "Please enter a URL" error message at submission - it sure seems worth it!

Exupery:
Changed the PR to use a normal `text` input and rely on the existing server side code to validate, resulting in the following behavior:
* submitting a link *without* a scheme (e.g. `example.com`) will result in successful link submission by relying on the [existing logic](https://github.com/reddit/reddit/blob/af09fa8dee69bef4f65a1662d9ad91c2329946e1/r2/r2/lib/utils/utils.py#L306-313) to add `http://` if no scheme is present (e.g. `http://example.com`)
* submitting an *invalid* URL with a scheme (e.g. `http://#$%@`) will display the existing [BAD_URL](https://github.com/reddit/reddit/blob/af09fa8dee69bef4f65a1662d9ad91c2329946e1/r2/r2/lib/errors.py#L36) error
* likewise, submitting a link *without* a scheme (e.g. `#$%@`) that does not become valid by adding a scheme will display the existing [BAD_URL](https://github.com/reddit/reddit/blob/af09fa8dee69bef4f65a1662d9ad91c2329946e1/r2/r2/lib/errors.py#L36) error
* mobile devices that display special keyboards when inputting URLs will now use the default keyboard - hopefully a day will come when browsers support the HTML5 [inputmode](http://www.wufoo.com/html5/attributes/23-inputmode.html) attribute, which would allow "hinting" they should use the URL keyboard, but without expecting validation
* valid submissions *with* a scheme are not affected by this PR

Boriss:
:fish: 
Sounds great!  tl;dr'ing above for URL input:
foo.com = valid, yay!
http//foo and foo = message: "you should check that url" (we need to improve that, but it's the same as current behavior)
mobile devices at input = no special URL-field input keys

wafflecopter05:
Couldn't you keep the textbox of type `url` and just change the `pattern` attribute to `.+`? I think it would be the same effect, while allowing mobile devices to still have special URL-field input keys.

Exupery:
@wafflecopter05 the `pattern` attribute provides *additional* validation, it doesn't supersede validation done by the browser for `url` or `email` inputs, e.g. [this fiddle](http://jsfiddle.net/gexq0y3h/)

bsimpson63:
Thanks for this contribution but it's been inactive for a while and I'm going to close it.

We hope to do some work on the submission flow in the future.

PR #1235, opened 2015-01-10
Wrap multi-reddit names in the search bar

paulgeringer:
Fixes an annoying visual discrepancy. Not too sure if there are other places where this should go. It appears to work in Chrome on OS X.

### DIFF
--- a/r2/r2/public/static/css/reddit.less
+++ b/r2/r2/public/static/css/reddit.less
@@ -1750,6 +1750,7 @@ body.with-listing-chooser.explore-page #header .pagename {
     margin: 5px 305px 5px 0px;
     border: 1px solid @info-border;
     font-size: small;
+    word-wrap: break-word;
 }
 
 .infobar img {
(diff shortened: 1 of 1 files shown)


bsimpson63:
Can you include a screenshot of before/after?

paulgeringer:
Before:
Imgur

After:
Imgur

bsimpson63:
This should probably use a more specific selector--maybe `#searchexpando label` will work. Also what do you think about doing "text-overflow: ellipsis;" instead of wrapping?

paulgeringer:
Works for me! Done.

r2/r2/public/static/css/reddit.less
```diff
@@ -2688,6 +2688,10 @@ textarea.gray { color: gray; }
     border-radius:3px;
 }
 
+#searchexpando label {
+    text-overflow: ellipsis;
```
bsimpson63: This doesn't seem to be sufficient.

```
display: block;
overflow: hidden;
text-overflow: ellipsis;
white-space: nowrap;
```
seems to get the behavior you want

bsimpson63:
:nail_care:

bsimpson63:
Closing. Comment here if you want to reopen this.

PR #1237, opened 2015-01-15
Losslessly reduced size of PNG images

KeenRivals:
/u/TyIzaeL here! I was able to losslessly reduce the size of PNG images in reddit from 1397KiB to 1192Kib (15% reduction) using optipng and advancecomp tools.


umbrae:
Hey TylzaeL, thanks for the PR. Sadly almost all of these are already going to be optimized on build when they're added to our spritesheet: (ex: https://www.redditstatic.com/sprite-reddit.png)

All images references with `/* SPRITE */` in our less file are automatically added to that sheet. I suspect most of the gains here will be lost after removing those sprited images, sadly. Going to close this as a result, but if you wanted to poke and see if there are non-sprited images that can use lossless optimization, I'd be happy to take a look at that PR for you.

PR #1238, opened 2015-01-15
Losslessly reduced non-sprite PNGs with Optipng and Advancecomp.

KeenRivals:
Coming back from #1237, only PNGs that were referenced in less files that were not commented as sprites were reduced. Size of the target PNGs was losslessly reduced from 483KiB to 365KiB for 118KiB savings (24%).

I should mention with the sprite, it would be possible to save about 1KiB by using advdef -z4 -i60 on it. But I'm not really sure how you would incorporate that into your less build process (I am not familiar with less) and given the gains are so small for that one file it may not be worth the effort.


xiongchiamiov:
> I should mention with the sprite, it would be possible to save about 1KiB by using advdef -z4 -i60 on it. But I'm not really sure how you would incorporate that into your less build process (I am not familiar with less)

The spriting stuff is all done in Python, actually, in [`lib/nymph.py`](https://github.com/reddit/reddit/blob/master/r2/r2/lib/nymph.py) (which [is run](https://github.com/reddit/reddit/blob/master/r2/Makefile#L165) by the Makefile).

KeenRivals:
So if you were willing to add advancecomp and optipng as dependencies in 
your makefile, you could use them to compress the static images after 
they are built. I belive recent versions of Ubuntu include the correct 
versions of advancecomp and optipng. IIRC, the version of advancecomp in 
Debian stable does not support the -i option, which gives you better 
savings.

umbrae:
https://github.com/reddit/reddit/commit/d6358c74840ee3189fa9f1def9f7d045fa323d74

Thanks for your efforts, @KeenRivals! I've also given you an open sourcerer award on reddit.

PR #1250, opened 2015-02-04
Change "liked" "disliked" to "upvoted" "downvoted" in prefs page

allthefoxes:
Reddit itself states that upvotes/downvotes should not be used as "I like/dislike this" buttons in the reddiquette, and changing this might be a good idea to both clear the meaning of the options, and more adhere to reddit's own guidelines.

Sorry for the anon aws user commit.

### DIFF
--- a/r2/r2/templates/prefoptions.html
+++ b/r2/r2/templates/prefoptions.html
@@ -162,11 +162,11 @@
           ${_("(such as the source subreddit or the content author's url/name)")}
         </span>
       </p>
-      <p>${checkbox(_("don't show links after I've liked them"), "hide_ups")}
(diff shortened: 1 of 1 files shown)


Deimos:
Thanks, allthefoxes. I've actually already got another pull request in progress internally that does this as well, but goes a little further and removes all usages of "liked" and "disliked" from the site (such as the tabs on the userpage). There's a couple small things that I still need to figure out to be able to deploy that one, but hopefully it'll be out fairly soon. I'm going to close this pull req since mine will take care of this as well.

allthefoxes:
I don't know how to make the emoji thumbs up, but if I did, I would put it right here: -

Alright. Thanks!

umbrae:
`:+1:`

PR #1252, opened 2015-02-07
Allow all users with https forced to use 2fa auth.

md-5:


Deimos:
If only it was this simple. This is only used when enabling "admin mode", so enabling it for non-admin users will not do anything.

md-5:
Haha I was just going to look into that.

PR #1255, opened 2015-02-10
Update authentication.py

grandpaslab:
Add support for automatic account creation when using basic auth.

### DIFF
--- a/r2/r2/lib/authentication.py
+++ b/r2/r2/lib/authentication.py
@@ -33,6 +33,7 @@ from pylons import g, c, request
 from urllib import unquote
 
 from r2.models import Account, NotFound
+from r2.models.account import register
 from r2.lib.utils import constant_time_compare, parse_http_basic
 from r2.lib.require import RequirementException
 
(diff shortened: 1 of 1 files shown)


grandpaslab:
I'm running a reddit instance on an intranet. My use case requires reddit account creation when user first logs in via basic auth, credentials being maintained elsewhere, in this case LDAP/AD. 

This change provides separate wrappers, 'http_basic', which maintains current functionality, and 'http_basic_autoregister' which creates a new account if one is not found for the given username.

spladug:
Way cool! So glad to hear someone's using this pluggable thing :)

grandpaslab:
Just updated this. Should not check password if using http_basic_autoregister. Passwords managed elsewhere.  

Probably http_basic should follow this pattern as well, but I'm not gonna mess with existing functionality.

spladug:
Thanks for your contribution! I've merged it as 1b6df45fa121aba1ffa438f74cd9e16c7afa3922 with some minor tweaks.  I've also taken the opportunity to pull the authentication provider into the modern provider system which should be a bit more flexible in the future. Please let me know your reddit username if you'd like an open sourcerer trophy.

grandpaslab:
Woohoo!  Thanks for this. Reddit username = github username = grandpaslab.

PR #1256, opened 2015-02-13
Add missing extensions to info endpoint in API docs

trevorsenior:
The endpoint [GET [**/r/**_subreddit_]**/api/info**](https://www.reddit.com/dev/api#GET_api_info) is missing the extensions `[ .json | .xml ]`

Discussed in https://github.com/trevorsenior/snoocore/issues/92

### DIFF
--- a/r2/r2/controllers/api.py
+++ b/r2/r2/controllers/api.py
@@ -173,7 +173,11 @@ class ApiController(RedditController):
         things=VByName('id', multiple=True, limit=100),
         url=VUrl('url'),
     )
-    @api_doc(api_section.links_and_comments, uses_site=True)
+    @api_doc(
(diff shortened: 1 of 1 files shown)


ThorArakis:
tbh - the .json version doesn't actually work as expected. I don't think we should be explicitly adding it to the documentation if it's not functional.

ThorArakis:
Your pr has sparked some internal discussion on how to better represent the fact that those endpoint support rss consumption (.xml) but that our api's really only support json. We will be continuing to support rss feeds so in order to make this a little more clear we've changed to a 'rss support' flag instead of the old syntax. [cf233d2d2595e3307e5ee85a03c1c8b6d56ce9bb]

PR #1258, opened 2015-02-15
fix sidebar width issues

jibberia:
Many subreddits don't test their sidebar content with custom CSS turned off, and they have content that is too wide, causing the entire document to scroll left and right. overflow-x: scroll will make just the sidebar scroll horizontally instead.


xiongchiamiov:
See also the discussion [here](https://www.reddit.com/r/bugs/comments/2vza1f/sidebar_css_bug_fix_included/).

umbrae:
From a product perspective, I don't think we ever want the sidebar to scroll, so we'd prefer it to look kinda broken for the subreddits so they're encouraged to fix it within their CSS/content itself.

Thank you for the contribution though!

PR #1263, opened 2015-02-24
Fix reddit/reddit-vagrant setup issues

trevorsenior:
Ubuntu images that vagrant pulls do not come shipped with curl. Ensure that curl is on the system before proceeding.

More privileges were needed to change the owner of the `developer.update` to $REDDIT_USER.

### DIFF
--- a/install-reddit.sh
+++ b/install-reddit.sh
@@ -104,6 +104,9 @@ export DEBIAN_FRONTEND=noninteractive
 # dependencies are found
 apt-get update
 
+# ensure that we have curl on this system
+apt-get install $APTITUDE_OPTIONS curl
+
 # add the reddit ppa for some custom packages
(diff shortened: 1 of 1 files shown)


ThorArakis:
Thanks for adding this. However for the time being, we decided to use wget (301bc90d19d1baa65f749efbec3c9224bc72ef61) in favor of adding a dependency on curl.

And in regards the the `sudo chown` issue. This will solve the problem with the initial vagrant provisioning, but it will cause the script to fail if it is run by hand from within the VM. We are currently trying to figure out a long term solution in regards to keeping the install-script free from vagrant specific code while still making sure that vagrant works. I'll update redditdev once we reach a decision.

trevorsenior:
Awesome thanks for the info.

PR #1266, opened 2015-02-26
Add default /etc/hosts configuration

tojrobinson:
Resolve `REDDIT_DOMAIN` to `127.0.0.1` by default via an `/etc/hosts` entry (if one doesn't already exist). Seems like a fairly standard/innocuous action.

A stronger condition `grep -q "$REDDIT_DOMAIN" ...` could be used; however, since the entry is appended to the bottom of the file, any existing entries for the domain will be preferred during resolution anyway.


spladug:
Hi there, this is only really useful if you're running your browser etc. on the same host as the reddit install isn't it? My understanding is most people use this with a VM so changing /etc/hosts would do nothing for their browser and might end up confusing as a result.

(personally, I install `avahi-daemon` then mDNS gives me `reddit.local` etc. for free)

tojrobinson:
Well it would depend where you spawn the browser from of course, yes. I'm using a VM as well but I open the browser from within the VM so I can leave my home environment alone altogether. You can just `apt-get install ubuntu-desktop` or use the desktop image to begin with since it is largely the same distro bar default packages (since 12.04 the kernel is the same I believe).

So, if you're running an encapsulated env, it will just work, otherwise you are in the same predicament as before. Confusing I guess if you read the whole script and expect the entry to work cross host.

I guess I interpreted the 'editing `/etc/hosts` on the host machine' from the install script docs to mean the host on which reddit is to run, i.e., inferring everything was on the same host. Maybe a misunderstanding.

spladug:
I'm sorry but I don't think we'll be merging this.  Hostname resolution is outside the scope of what the install script can/should know. It may be nice to add some documentation to the wiki about how to set it up in various environments though.

PR #1269, opened 2015-03-01
API: Add support for JSON responses from /api/subreddit_stylesheet

ArrestedDevelopment:
Previously, the only response one could get from reddit when submitting a stylesheet is the difficult-to-parse one intended for reddit's jQuery pseudo-eval library.

This commit also adds "modhash" to the response from `/api/upload_sr_img.json`, and adds "url" to the response from `/about/stylesheet.json`; Additionally, the `.json` variants of `/api/subreddit_stylesheet` and `/api/upload_sr_img.json` have been added to the docs as they're the only way to use these methods while in a browser userscript. (that I've found, at least)

### DIFF
--- a/r2/r2/controllers/api.py
+++ b/r2/r2/controllers/api.py
@@ -71,6 +71,10 @@ from r2.lib.pages import (
     WikiMayContributeTableItem,
 )
 
+from r2.lib.jsontemplates import (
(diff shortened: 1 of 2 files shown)


ArrestedDevelopment:
Going to close this - and then redo this in a new pull request

PR #1270, opened 2015-03-01
Make date format localizable after "this post was submitted on"

noamraph:
Current translation doesn't affect thing.datefmt (it's always "%d %b %Y"), so I changed it so that the format string is translated immediately.

### DIFF
--- a/r2/r2/lib/template_helpers.py
+++ b/r2/r2/lib/template_helpers.py
@@ -32,6 +32,7 @@ from r2.lib.static import static_mtime
 from r2.lib import js
 
 import babel.numbers
+import babel.dates
 import simplejson
 import os.path
 from copy import copy
@@ -590,6 +591,11 @@ def format_number(number, locale=None):
 
 def format_percent(ratio, locale=None):
     return babel.numbers.format_percent(ratio, locale=locale or c.locale)
+
+
+def format_date(date, locale=None):
+    return babel.dates.format_date(date, _("dd MMM yyyy"),
+                                   locale=locale or c.locale)
 
 
 def html_datetime(date):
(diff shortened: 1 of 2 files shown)


r2/r2/templates/linkinfobar.html
```diff
@@ -39,7 +39,7 @@
       &#32;
     </span>
     <time datetime="${html_datetime(thing.a._date)}">
-      ${thing.a._date.strftime(thing.datefmt)}
+      ${thing.a._date.strftime(_('%d %b %Y'))}
```
bsimpson63: something from babel might be more appropriate http://babel.pocoo.org/docs/api/dates/

noamraph: Do you mean something like

    format_date(thing.a._date, format=_('dd MMM yyyy'), locale=_('en_US'))

How do I get the locale name?

spladug: `c.locale` should contain the current locale. See [other users of the babel functions for examples](https://github.com/reddit/reddit/blob/master/r2/r2/lib/template_helpers.py#L586-L590).

r2/r2/lib/template_helpers.py
```diff
@@ -592,6 +593,11 @@ def format_percent(ratio, locale=None):
     return babel.numbers.format_percent(ratio, locale=locale or c.locale)
 
 
+def format_date(date, locale=None):
+    return babel.dates.format_date(date, _("dd MMM yyyy"),
```
bsimpson63: Why is the format string marked for translation?

umbrae:
Closing due to idleness, feel free to reopen if you intend to continue working on it.

PR #1271, opened 2015-03-01
Use <bdi></bdi> HTML tags to fix ordering in RTL direction

noamraph:
Without those, instead of "(1) user" you get "(user (1"
Instead of "user[S]" you get "[user[S".


umbrae:
Sadly we don't yet support RTL - when we focus more on international though we will take a look at RTL as a whole. Thanks for the contribution.

PR #1272, opened 2015-03-01
Make "save" and "cancel" button labels translatable

noamraph:


umbrae:
We're actually looking at changing these terms based on context right now which moves this around a bit. I'll make it translatable if we move on it, thanks for the reminder!

PR #1273, opened 2015-03-01
Use a format string for "3 days ago", don't force"ago" after "3 days"

noamraph:


r2/r2/lib/template_helpers.py
```diff
@@ -607,11 +607,11 @@ def simplified_timesince(date, include_tense=True):
     if date > timeago("1 minute"):
         return _("just now")
 
-    since = []
-    since.append(timesince(date))
+    since = timesince(date)
     if include_tense:
-        since.append(_("ago"))
-    return " ".join(since)
+        return _("%s ago") % since
```
bsimpson63: I think if you do this like `_("%(time)s ago")` it will be clearer for the translators.

kemitche: "%s ago" is already translated (it exists as a source in the wiki code); so in this case, I'd say it's actually better off with the original implementation...

bsimpson63:
Thanks, I'll get this merged.

:fish:

bsimpson63:
Also in light of @kemitche's comment I'll just merge the initial commit so we don't have to translate a new string.

bsimpson63:
Merged as 531fc3e32ed22a4f70a545ca52077ee78bae5a0b. Sorry it took so long.

If you tell me your reddit username I'll give you a trophy.

PR #1274, opened 2015-03-02
Follow current SEO best practices

jedberg:
When I wrote this function, the best practice was to use an underscore, but now a hyphen is the best practice.

The beauty of this is that since what's actually written in that part of the URL is ignored by the middleware, this change is backwards compatible despite the fact that it will change how every past title is rendered.

### DIFF
--- a/r2/r2/lib/utils/utils.py
+++ b/r2/r2/lib/utils/utils.py
@@ -936,19 +936,19 @@ rx_underscore = re.compile('_+', re.UNICODE)
 def title_to_url(title, max_length = 50):
     """Takes a string and makes it suitable for use in URLs"""
     title = _force_unicode(title)           #make sure the title is unicode
-    title = rx_whitespace.sub('_', title)   #remove whitespace
+    title = rx_whitespace.sub('-', title)   #remove whitespace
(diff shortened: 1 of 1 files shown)


Deimos:
I think this is a good idea for an update, but this pull won't work correctly as-is. For example, the `rx_notsafe` replacement will remove all the hyphens added by the previous whitespace replacement, since `\W` matches hyphens. I'll open a fixed version of the pull internally to see if this is something we want to do (and see if there are any other changes we should make at the same time).

Thanks for the idea, this is probably long overdue.

empyrical:
When this change makes it into reddit, you should give subreddit mods a day or two's notice, as a lot of stylesheets (like the one on /r/TodayILearned, as an example) use these submission URLs in some of their CSS selectors.

jedberg:
The subreddit mods should probably stop doing that, since any arbitrary text in that field is valid.  For example, this totally works:

http://www.reddit.com/r/todayilearned/comments/3d57fu/i-can-put-*_anything_*-i-want-here/

(real url is http://www.reddit.com/r/todayilearned/comments/3d57fu/til_that_the_type_ab_personality_theory_was/)

PR #1275, opened 2015-03-03
Solr search

grandpaslab:
Shim to point search at a Solr instance.

### DIFF
--- a/r2/example.ini
+++ b/r2/example.ini
@@ -368,16 +368,16 @@ CLOUDSEARCH_SUBREDDIT_SEARCH_API =
 CLOUDSEARCH_SUBREDDIT_DOC_API =
 #
 # endpoint for link search 
-SOLR_SEARCH_HOST = 
+SOLR_SEARCH_HOST = 127.0.0.1
 # endpoint for link upload
-SOLR_DOC_HOST = 
+SOLR_DOC_HOST = 127.0.0.1
 # endpoint for subreddit search
-SOLR_SUBREDDIT_SEARCH_HOST = 
+SOLR_SUBREDDIT_SEARCH_HOST = 127.0.0.1
 # endpoint for subreddit upload
-SOLR_SUBREDDIT_DOC_HOST = 
+SOLR_SUBREDDIT_DOC_HOST = 127.0.0.1
 # solr port
 SOLR_PORT = 8983
-# solr core name (default: 'collection1')
+# solr core name (defaults to 'collection1' if empty)
 SOLR_CORE = 
 # default batch size is 500
 # limit is hard-coded to 1000
--- a/r2/example.ini
+++ b/r2/example.ini
@@ -353,6 +353,11 @@ wiki_max_page_separators = 3
 
 
 ############################################ SEARCH
+#
+# search type defaults to cloudsearch
+# set to 'solr' to use a solr index
+SEARCH_TYPE = 
+#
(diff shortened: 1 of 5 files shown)


kemitche:
High level comment: Can you refactor the `SEARCH_TYPE` toggle to follow our "provider" pattern? You can use the media_provider as an example; see line ~430 of `app_globals.py` and the code under `r2/r2/lib/providers/media` for examples on how that works, but the short of it is you have a base class that defines the interface, and subclasses that define required config and the implementation of the provider.

r2/r2/lib/solrsearch.py
```diff
@@ -0,0 +1,564 @@
...
+from lxml import etree
+from r2.lib import amqp, filters
+from r2.lib.db.sorts import epoch_seconds
+import r2.lib.utils as r2utils
+from r2.models import (Account, Subreddit, All, DefaultSR,
+                       MultiReddit, DomainSR, Friends, ModContribSR,
+                       FakeSubreddit)
+from pylons import g, c
+
+from r2.lib.cloudsearch import *
```
kemitche: Please don't use `import *`; it can cause subtle import bugs (such as shadowing the `datetime` class with the `datetime` module).

r2/r2/lib/solrsearch.py
```diff
@@ -0,0 +1,564 @@
...
+    def _get_sr_restriction(sr):
+        '''Return a solr-appropriate query string that restricts
+        results to only contain results from self.sr
+        
+        '''
+        bq = []
+        if (not sr) or sr == All or isinstance(sr, DefaultSR):
+            return None
+        elif isinstance(sr, MultiReddit):
+            bq = ["(or"]
```
kemitche: Is the query generated here valid for Solr? This code looks unchanged from cloudsearch which has a fairly specific syntax.

r2/example.ini
```diff
@@ -361,7 +366,22 @@ CLOUDSEARCH_DOC_API =
 CLOUDSEARCH_SUBREDDIT_SEARCH_API =
 # endpoint for subreddit upload
 CLOUDSEARCH_SUBREDDIT_DOC_API =
-
+#
+# endpoint for link search 
+SOLR_SEARCH_HOST = 127.0.0.1
```
kemitche: I'm a wary of adding unused config that isn't enabled by install-reddit.sh. My preference would be to leave these values unset in example.ini and document them appropriately here or in the provider implementation.

solr/solrconfig.xml
```diff
@@ -0,0 +1,1905 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
kemitche: Is this file necessary, or would installing Solr provide a similar default file? If there aren't any reddit specific customizations in it, please omit it.

grandpaslab:
Will do.

kemitche:
A possibility for you to consider: if the search code can be properly provider-ized, then the Solr-specific pieces could be contained in a separate reddit-plugin on a github repo you control; that would then mean you could modify it as you see fit rather than requiring merging changes.

kemitche:
:nail_care: (First round of feedback complete, let me know when you're ready for round 2 of review or get in touch with me on reddit if you want to further discuss/clarify anything)

kemitche:
Closing this pull request as I believe it's been superceded by #1289

PR #1282, opened 2015-03-15
CSS Filter: allow property: clip-path. allow functions: circle, ellipse, etc.

GusGold:
`clip-path`: Used to clip an element via inside of polygon. https://developer.mozilla.org/en-US/docs/Web/CSS/clip-path

`circle`, `ellipse`, `line`, `polygon`, `polyline`: Functions used to create SVG shapes. Same uses as `rect`. https://developer.mozilla.org/en-US/docs/Web/SVG/Element/polygon


spladug:
Thanks! Merged as 5641565b02b33a70133531d473036b0506385a61.

Let me know your reddit username if you'd like an open sourcerer trophy.

GusGold:
@spladug /u/GusGold

Cheers

PR #1287, opened 2015-03-29
Add support for text-fill-color

gavin19:
Currently a webkit-only property.


JordanMilne:
Looks like this has been around for a while and still isn't on a standards track. Closing this in the interest of keeping stylesheets mostly cross-browser compatible.

PR #1290, opened 2015-04-06
Add SVG support to custom images

huw:
SVG support for subreddit stylesheets allows reddit to move forward into the future of high-resolution displays, because scalability is key.

In making this change, there are a number of security issues that need to be considered with relation to SVG and XML files that may be uploaded to a server, especially one the size of reddit. As such, this change employs a whitelist of allowable tags and attributes (much like the CSS validator), and takes all necessary steps to make the SVGs are secure as possible.

The program, when tested on [my own server](http://50.116.12.143/), correctly sanitised each and every vulnerability on the [HTML5Sec page](https://html5sec.org/#svg) that was listed under SVG _and_ XML. That being said, I would be happy to aid any additional security testing in any way that I can.

I also have a wiki page ready, which would outline the best practices for submitting an SVG file, because there are some nuances which could involve sacrificing SVG functionality that might confuse the user. The draft is at [this link](https://gist.github.com/huw/0cfd95695c2829f372ea).

Looking forward to hearing from you guys, this is a change which I really hope we can implement 😊

### DIFF
--- a/r2/r2/lib/media.py
+++ b/r2/r2/lib/media.py
@@ -35,6 +35,8 @@ import urllib2
 import urlparse
 import gzip
 
+from xml.etree import ElementTree
+
 import BeautifulSoup
 import Image
 import ImageFile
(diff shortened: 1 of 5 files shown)


r2/r2/lib/media.py
```diff
@@ -86,6 +303,36 @@ def str_to_image(s):
...
+            elif units == 'mm':
+                dimensions[i] = num * 3.543307
+            elif units == 'cm':
+                dimensions[i] = num * 35.43307
+            elif units == '%':
+                dimensions[i] = -num / 100.0
+            else:
+                dimensions[i] = num
+    except KeyError:
+        dimensions = [500, 500]
```
JordanMilne: Any particular logic to these defaults? What if `width` is specified and very large, but `height` is unspecified or vice-versa? Can other stuff like a `style` on the root `svg` tag affect the effective dimensions?

r2/r2/lib/media.py
```diff
@@ -205,6 +452,83 @@ def optimize_jpeg(filename):
...
         subprocess.check_call(("/usr/bin/jpegoptim", filename), stdout=devnull)
 
+class _SVGSoup(BeautifulSoup.BeautifulStoneSoup):
+    # Fixes unexpected conversion from `<g><g></g></g>` into `<g></g><g></g>`
+    nests = ('defs', 'g', 'marker', 'mask', 'pattern', 'svg', 'symbol', 'use')
+    NESTABLE_TAGS = BeautifulSoup.buildTagMap([], nests)
+
+def _sanitize_svg(image):
+    elements = [x.lower() for x in _SAFE_SVG_ELEMENTS]
+    attributes = [x.lower() for x in _SAFE_SVG_ATTRIBUTES]
```
JordanMilne: These var names are kind of misleading, maybe `element_whitelist` and `attribute_whitelist`?

r2/r2/lib/media.py
```diff
@@ -205,6 +452,83 @@ def optimize_jpeg(filename):
...
+
+    # BS3 gets confused with some self-closing tags. We can safely 
+    # convert them with regex and afterward, we strip out unwanted <!> 
+    # and <?> tags.
+    #
+    # Keep in mind that `[^>]` goes in place of `.`, because it matches
+    # newline chars as well. One could also flag re.DOTALL, but we need
+    # to also not match `>`s.
+    image = re.sub(r'<([^>\s]+)\s*([^>]*)/>', r'<\1 \2></\1>',
+                   re.sub(r'<![^>]*>|<\?(?!xml)[^>]*>', r'', image))
```
JordanMilne: The regex gymnastics before parsing strikes me as weird, is there a parser that'd be more appropriate here than `BeautifulStoneSoup`? If I were writing this from scratch I'd use one of the `defusedxml` parsers.

r2/r2/lib/media.py
```diff
@@ -205,6 +452,83 @@ def optimize_jpeg(filename):
...
+                         elem[attr].startswith('#')) == False):
+
+                    # BS3 lowercases attribute names - this can be bad
+                    # Using the two arrays we made, we can revert this
+                    # (This is also why we delete every attribute first)
+                    new_attr = _SAFE_SVG_ATTRIBUTES[attributes.index(attr)]
+                    to_save[new_attr] = elem[attr]
+
+            for key in to_del:
+                del elem[key]
```
JordanMilne: It'd be better to create a version of the tree with only known-clean attrs and elems, rather than mutate the existing tree. That way you're not vulnerable to weird issues if whatever you're using to walk the tree misses something security-relevant (a doctype decl, a PI, a comment, etc.)

r2/r2/lib/media.py
```diff
@@ -205,6 +452,83 @@ def optimize_jpeg(filename):
...
+                value = attribute_set[1]
+                to_del.update({attr})
+
+                # When safe:
+                # Expression 1 == True
+                # Expression 2 == None
+                # Expression 3 == False
+
+                if (attr in attributes and
+                    re.search(r'url\((?!#)', elem[attr]) == None and
```
JordanMilne: There's a few ways around the `url(` check. Off the top of my head, Firefox lets you do things like `r"\u\r\l("`, and I seem to remember one browser allowing embedded nulls lik `b"ur\x00l("`.

r2/r2/lib/media.py
```diff
@@ -205,6 +452,83 @@ def optimize_jpeg(filename):
...
+            for key in to_save:
+                elem[key] = to_save[key]
+
+        elem.name = _SAFE_SVG_ELEMENTS[elements.index(elem.name)]
+
+    # Remove repeated <?xml?> tags, then minify
+    xml = re.sub(r'(<\?xml.*>)\1+', r'\1',
+                 re.sub(r'[\t\r\n] *', '', soup.prettify("utf-8")))
+
+    ElementTree.fromstring(xml) # Will raise error if XML is broken
```
JordanMilne: In what cases is the XML going to fail to parse after serialization?

r2/r2/lib/media.py
```diff
@@ -205,6 +452,83 @@ def optimize_jpeg(filename):
...
+            elem.decompose()
+            continue
+
+        if elem.attrs:
+            to_del = set()
+            to_save = dict()
+
+            for attribute_set in elem.attrs:
+                attr = attribute_set[0]
+                value = attribute_set[1]
```
JordanMilne: Can this be `for attr, value in elem.attrs:`?

JordanMilne:
Hey, security engineer at @reddit here! 

Thanks for the PR! I'd really like to implement this, but honestly, I'm kind of iffy on allowing arbitrary SVG uploads. There isn't currently anywhere else in the app that we deal with completely user-controlled documents. Even in `souptest.py` we make sure we control the start of the document and that there are no inline styles, even though everything passed into that is semi-trusted.

I don't think any of us are familiar enough with sanitizing arbitrary SVG docs and the various browser implementations to be sure we're not opening up a huge info leak, either. Just verifying that our current CSS validator had no possible bypasses took a week of constantly poking around in VMs with different browsers.

r2/r2/lib/media.py
huw: The [500,500] is there because it's more complicated to actually run through each tag and calculate the overall width and height. It's completely arbitrary.

For the rest of that expression, it looks like the values are working for a screen at 90 ppi, although I did admittedly borrow these from someone else's script. The chances of someone using a unit other than 'px' is very low, though.

I think it's probably easier, rather than trying to implement every possible use-case, to make the user do the work here. All of the big SVG generators (Sketch, Inkscape, Illustrator) will specify width/height in px, so if there is an issue like an unspecified height, the user has probably crafted their tags from scratch.

r2/r2/lib/media.py
huw: I chose to use BS3 because it was already imported and was able to do _most_ of the job I wanted. I would be happy to use `defusedxml.lxml`, but I wasn't really sure what the policy was on adding new packages -- nor how exactly to do it.

r2/r2/lib/media.py
huw: I found this happened when I tried to nest <!ENTITY> tags within a <!DOCTYPE> (for billion laughs or similar XML exploits). I wasn't sure if there were any other instances so I kept this in (not a security expert!)

r2/r2/lib/media.py
```diff
@@ -205,6 +452,83 @@ def optimize_jpeg(filename):
...
+                del elem[key]
+
+            for key in to_save:
+                elem[key] = to_save[key]
+
+        elem.name = _SAFE_SVG_ELEMENTS[elements.index(elem.name)]
+
+    # Remove repeated <?xml?> tags, then minify
+    xml = re.sub(r'(<\?xml.*>)\1+', r'\1',
+                 re.sub(r'[\t\r\n] *', '', soup.prettify("utf-8")))
```
huw: The duped XML was something I didn't get with BS4, but only with BS3. It occurs when I used a tag up the top of the document like `<?xml-stylesheet?>` (no clue why). I presume this wouldn't happen if lxml were used.

huw:
Thanks for looking at it! I am definitely on your side in that SVG is very difficult to implement securely, especially because in general it's such a bloated and weird format -- let alone its use of XML. I've replied to most of your comments above, and for the others, I left them blank because I pretty much agree with those suggestions.

I agree that rewriting this to be properly and fully secure (with `lxml` or a different XML library) is probably the best way forward. I don't think as a humble designer/programmer I could do this as well as a proper security engineer -- but I understand that implementing this will be tough and fragile to handle.

I would like to see a reddit which somehow supports vector graphics in the future, even if its in a highly limited capacity (like blanket banning `style` attributes) that requires the user to do the extra work. Consider this PR, I guess, as a kind of proof-of-concept mixed in with a call to action.

If you do end up going forward with this (please! 🙏), I did a fair bit of research before writing this. There's not much on it all, but I found [a paper on SVG exploits](https://www.ei.rub.de/media/hgi/veroeffentlichungen/2011/10/19/svgSecurity-ccs11.pdf), and a [frustratingly closed-source sanitiser](http://heideri.ch/svgpurifier/SVGPurifier/index.php) which emerged from the research. I don't know how big your security team is, or how much making this change is worth, but reddit is seriously an atrocity on retina devices, and I imagine that moving into the future things will only get worse for them. For now, though, I guess I can just use the usual `@media only screen and (min--moz-device-pixel-ratio: 2), only screen and (-o-min-device-pixel-ratio: 2/1), only screen and (-webkit-min-device-pixel-ratio: 2), only screen and (min-device-pixel-ratio: 2)` 😄

ysangkok:
MediaWiki has a SVG "verifier", as they call it: [includes/upload/UploadBase.php:checkSvgScriptCallback](https://github.com/wikimedia/mediawiki/blob/84899f991ff44f36b7359f0bbf8de2955b81a904/includes/upload/UploadBase.php#L1296)

JordanMilne:
Sorry but we're unlikely to add SVG support in the near future for the reasons described further up in the thread and in https://www.reddit.com/r/ideasfortheadmins/comments/37qaho/svg_support_for_subreddit_designs/crprpvx

PR #1294, opened 2015-04-11
Added AutoModerator to the config permission tooltip

Jakeable:
The config permission now has the ability to manage AutoModerator, so the tooltip should reflect this.

### DIFF
--- a/r2/r2/lib/app_globals.py
+++ b/r2/r2/lib/app_globals.py
@@ -561,7 +561,7 @@ class Globals(object):
         # the main memcache pool. used for most everything.
         memcache = CMemcache(
             self.memcaches,
-            min_compress_len=50 * 1024,
+            min_compress_len=1400,
             num_clients=num_mc_clients,
             binary=True,
         )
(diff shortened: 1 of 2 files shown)


xiongchiamiov:
The change itself is simple enough, but as for whether it's correct, probably best to :eyeglasses: @Deimos .  He's a bit occupied at the moment, so it may be a little while before he can pop in.

Deimos:
I think that's probably correct, though it might be a little better to say something more specific like "automoderator configuration". Mods need to have the "wiki" permission to be able to view the automod configuration, and the "config" one to be able to edit it.

Jakeable:
@deimos how does "AutoModerator config" sound? "Configuration" seems really long compared to the rest of the tooltip.

Deimos:
Sure, that should be good.

Jakeable:
Think this one is ready for primetime now?

Deimos:
Looks good to me, I'll get it added in. Thanks. :fish:

Deimos:
And deployed now: https://github.com/reddit/reddit/commit/e772017d2e5f51a55e9fe542b657f97b75233faa

Jakeable:
Thanks! Would it be possible to get the Open Sorcerer? :wink:

Deimos:
Hmm, I'm actually not sure if we give that for text updates or just actual functional code changes. I'll check with some of the others and confirm.

PR #1295, opened 2015-04-12
Added "theme-color" meta element to layout

robin0van0der0v:
Added a "theme-color" meta element to base layouts.
This will make the address and status bar bluish on Chrome in Android Lollipop.


umbrae:
I think long term we're not looking for lots of mobile traffic on desktop, so this probably is just extra markup without a lot of gain. Thanks for the contribution but I think we'll pass for now.

PR #1296, opened 2015-04-17
Update automoderator.py

Umdlye:
Add an invisible link to the AM disclaimer so it can be styled with

a[href="#AM"]~em

### DIFF
--- a/r2/r2/lib/automoderator.py
+++ b/r2/r2/lib/automoderator.py
@@ -79,7 +79,7 @@ if g.automoderator_account:
 else:
     ACCOUNT = None
 
-DISCLAIMER = "*I am a bot, and this action was performed automatically. Please [contact the moderators of this subreddit](/message/compose/?to=/r/{{subreddit}}) if you have any questions or concerns.*"
+DISCLAIMER = "[](#AM) *I am a bot, and this action was performed automatically. Please [contact the moderators of this subreddit](/message/compose/?to=/r/{{subreddit}}) if you have any questions or concerns.*"
 
 rules_by_subreddit = {}
 


13steinj:
@Umdlye This isn't needed. In fact, this makes things worse. The method you describe will cause any user who knows about this who types `[](#AM)` into their comment followed by Italics to also gain that styling.

But it's not needed regardless. A quick look at the html shows this:

     <div id="siteTable_t3_3emt5j" class="sitetable nestedlisting">
     (some non important code)
        <div class="entry unvoted">
            <p class="tagline">
                <a href="javascript:void(0)" class="expand" onclick="return togglecomment(this)">[–]</a>
                <a href="https://www.reddit.com/user/AutoModerator" class="author moderator may-blank id-t2_6l4z3" >AutoModerator</a>
                <span class="userattrs">
                    [
                    <a class="moderator" title="moderator of /r/13steinj, speaking officially" href="/r/13steinj/about/moderators">M</a>
                    ]
                </span>
                <span class="score dislikes">0 points</span><span class="score unvoted">1 point</span>
                <span class="score likes">2 points</span> 
                <time title="Sun Jul 26 05:13:41 2015 UTC" datetime="2015-07-26T05:13:41+00:00" class="live-timestamp">just now</time> 
                <a href="javascript:void(0)" class="numchildren" onclick="return togglecomment(this)">(0 children)</a>
            </p>
            <form action="#" class="usertext" onsubmit="return post_form(this, 'editusertext')" id="form-t1_ctgez7cx6w">
                <input type="hidden" name="thing_id" value="t1_ctgez7c"/>
                    <div class="usertext-body may-blank-within md-container ">
                    <div class="md">
                        <p>Your submission was automatically removed because self is not an approved site.</p>
    
                         <p><em>I am a bot, and this action was performed automatically. Please <a href="/message/compose/?to=/r/13steinj" rel="nofollow">contact the moderators of this subreddit</a> if you have any questions or concerns.</em></p>
                    </div>
    ......(more non important code here)

Knowing this, one can style the disclaimer via  knowing the comment's t3_ code. However this isn't ideal, as it doesn't apply to all Automod comments. Now, I wouldn't be able to alter reddit's code itself, but what you would need to do is somehow get all divs with a class of `.entry` and whose child is `.tagline a[href$="/AutoModerator"]` to have their own class or id of `automod-comment`

Deimos:
I'm going to close this pull req. I'll have to figure out if this is something we want to support, but I definitely agree with @13steinj that using an invisible link like this isn't a great approach for it. If it is something we want to do we can probably figure out a cleaner way since AutoMod's integrated into the site, so we might be able to add a special class to its comments or something.

PR #1302, opened 2015-04-26
Allow links with FQDN's

ysangkok:
See http://www.reddit.com/r/bugs/comments/33c6fx/domains_not_ending_in_cant_get_posted_breaks_tld/

### DIFF
--- a/r2/r2/tests/unit/lib/utils_test.py
+++ b/r2/r2/tests/unit/lib/utils_test.py
@@ -95,6 +95,32 @@ class UtilsTest(unittest.TestCase):
             utils.coerce_url_to_protocol('//example.com/foo', 'https'),
             'https://example.com/foo')
 
+    def test_sanitize_url(self):
+        self.assertEquals(
+            utils.sanitize_url('http://dk./'),
+            'http://dk/'
+        )
+
+        self.assertEquals(
+            utils.sanitize_url('http://google.com./'),
+            'http://google.com/'
+        )
+
+        self.assertEquals(
+            utils.sanitize_url('http://google.com/'),
+            'http://google.com/'
+        )
+
+        self.assertEquals(
+            utils.sanitize_url('https://github.com/reddit/reddit/pull/1302'),
+            'https://github.com/reddit/reddit/pull/1302'
+        )
+
+        self.assertEquals(
+            utils.sanitize_url('http://dk../'),
+            None
+        )
+
 
 class TestCanonicalizeEmail(unittest.TestCase):
     def test_empty_string(self):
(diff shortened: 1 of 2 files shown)


xiongchiamiov:
Thanks!

:eyeglasses: @JordanMilne , do you have any feedback on this?

JordanMilne:
does the URL that gets used on the site still have the trailing `.`, or does it get stripped?

r2/r2/tests/unit/lib/utils_test.py
```diff
@@ -95,6 +95,22 @@ def test_coerce_url_to_protocol(self):
             utils.coerce_url_to_protocol('//example.com/foo', 'https'),
             'https://example.com/foo')
 
+    def test_sanitize_url(self):
+        self.assertEquals(
+            utils.sanitize_url('http://dk./'),
+            utils.sanitize_url('http://dk/')
```
xiongchiamiov: I think it's probably a better idea to have a static string to compare against - that is, have one of the arguments to `assertEquals` be just `'http://dk/'` (I think? whatever the result is), and have a separate test case covering the more standard inputs.

ysangkok:
@JordanMilne The dot is stripped, all this patch does is remove the dot if it is there. It means that it won't be possible to link sites that rely on the dot, but then again, they are not postable now either.

xiongchiamiov:
:nail_care: (see [The Code Salon](https://github.com/reddit/reddit/wiki/The-Code-Salon))

ysangkok:
Is there any chance of getting this merged? If there is, I could apply the changes suggested by @xiongchiamiov

13steinj:
@ysangkok The nailcare means that the collaborator either wants a response, or a change in the code that you made. I'm guessing @xiongchiamiov wanted that simple change made, and afterwards, you'd notify him with `:haircut:` that you made the change.

Pokechu22:
@ysangkok: The emoji that was there means "I've finished my review pass and am awaiting your responses", so I would say yes, you should make those requested changes.  Once you've made them, reply with a `:haircut:`.  [More info here](https://github.com/reddit/reddit/wiki/The-Code-Salon); it's a somewhat silly system but it's what they use.

ysangkok:
@xiongchiamiov :haircut:

13steinj:
@ysangkok Given the fact that there are literally a million commits; I think you did something wrong, but since I don't know what I can't give you a git command to fix it.

ysangkok:
@13steinj sorry about that, I took care of it.

13steinj:
@ysangkok No need for sorry, just wanted to let you know :P

xiongchiamiov:
Cool, looks good to me. :fish: It's a bit late today to push this out, and we try to not deploy on Fridays unless we have a good reason to, so I'll try to get this out on Monday for you.

xiongchiamiov:
1749dd9275202e0b4710b4697f4441539c8cca08

PR #1303, opened 2015-04-26
shorten GeoIP.open calls using map

Matthew94:
instead of doing the same try/except block three times, use map with a function that contains the block


bsimpson63:
Thanks for this pull request, but we won't be accepting it. Again, your changes look correct, but performance isn't an issue here and I find your changes to be less readable.

Sorry!

Matthew94:
No worries.

Thanks for taking a look.

PR #1305, opened 2015-04-26
shortened iterators using gen exps

Matthew94:


bsimpson63:
Thanks for this pull request, but we won't be accepting it. While your changes look correct, performance is not an issue for this area of the code and I'd rather not risk breaking anything. If we had better (or any!) test coverage here it would be a different story.

Sorry!

Matthew94:
Fair enough then, thanks for taking a look at it.

PR #1306, opened 2015-04-29
Added link to reddit Enhancement Suite under the apps & tools section in the footer.

KamranMackey:
RES is an amazing add-on for reddit, and I think it would be awesome if it was in the apps & tools section in the footer.


xiongchiamiov:
While RES is pretty cool, we have limited footer space and thus need to limit that section only to official apps and tools; if we included RES, we'd also need to include many, many other browser extensions, mobile apps, irc bots, statistics sites, etc. to avoid showing favoritism.

Aside from being fair, we also don't want to mistakenly imply that we provide support for RES; we have our hands full with our own bugs, and the RES team does a great job with their support separately.

Thanks!

PR #1309, opened 2015-05-01
Perserve the originally attempted title when resubmitting a link.

Pokechu22:
Hello!  /u/pokechu22 here.  I've added support for saving the current title when resubmitting a link, in the same way that 1c8a271 did with the full URL.

### DIFF
--- a/r2/r2/controllers/front.py
+++ b/r2/r2/controllers/front.py
@@ -287,7 +287,8 @@ class FrontController(RedditController):
         infotext = None
         if request.GET.get('already_submitted'):
             submit_url = request.GET.get('submit_url') or article.url
-            resubmit_url = Link.resubmit_link(submit_url)
+            submit_title = request.GET.get('submit_title') or ""
+            resubmit_url = Link.resubmit_link(submit_url, submit_title)
             if c.user_is_loggedin and c.site.can_submit(c.user):
                 resubmit_url = add_sr(resubmit_url)
             infotext = strings.already_submitted % resubmit_url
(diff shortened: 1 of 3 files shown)


r2/r2/controllers/front.py
```diff
@@ -286,7 +286,8 @@ def GET_comments(
         infotext = None
         if request.GET.get('already_submitted'):
             submit_url = request.GET.get('submit_url') or article.url
-            resubmit_url = Link.resubmit_link(submit_url)
+            submit_title = request.GET.get('submit_title') or article.title
```
bsimpson63: Is it really appropriate to use `article.title` as the fallback? Wouldn't it be better to not pre-fill a title at all?

Pokechu22: I can see some use, although it doesn't seem too common.

If someone wants to resubmit an article, this way it would automatically supply the title as well.  But I see this as being very rare.  I feel like it could go either way.

bsimpson63: I think it's be less confusing to not per-fill a title at all. Could you make that change?

Pokechu22:
:haircut:

bsimpson63:
Thanks, this looks good. I'm going to test it a bit and then will get it merged.

bsimpson63:
Merged as c942cef02c71d3228cdaee234d3195aeec37e70e

What's your reddit username so I can give you a trophy?

Pokechu22:
[/u/pokechu22](https://reddit.com/u/pokechu22).

PR #1310, opened 2015-05-03
Reword some strings.

minecast:


allthefoxes:
A lot of these have no real point behind it except for maybe the last one (on gold) - The way they are currently for the rest make perfect sense, and you did not explain your reasoning as to why they should be changed.

umbrae:
I tend to agree. Thanks for your time, but I'm not sure these are a valuable addition. If you'd like to reword something in the future, a little reasoning would go a long way.

PR #1315, opened 2015-05-13
Hyphenate "autorenewing"`

473b:
The word "autorenewing" should be hyphenated.


bsimpson63:
Thanks for this pull request, but we're not going to accept it because it's not worth invalidating our translations for this minor correction.

Sorry!

PR #1316, opened 2015-05-13
Fix bug where literal HTML is placed into page.

justcool393:
/r/bugs post:
http://www.reddit.com/r/bugs/comments/35upvy/

### DIFF
--- a/r2/r2/controllers/error.py
+++ b/r2/r2/controllers/error.py
@@ -229,7 +229,7 @@ class ErrorController(RedditController):
             elif code == 500:
                 randmin = {'admin': random.choice(self.admins)}
                 failien_url = make_failien_url()
-                sad_message = safemarkdown(rand_strings.sadmessages % randmin)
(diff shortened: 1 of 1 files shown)


JordanMilne:
Thanks for the PR!

So the code here is actually correct. The issue is that some translations didn't get the markdown version of the error messages pulled in when that was converted to markdown. 

For example, the [Brazilian Portuguese translation](https://github.com/reddit/reddit-i18n/blob/25a519951278fa161cde510bc14222ef8c4faa8c/reddit_i18n/pt_BR/LC_MESSAGES/r2.po#L1700-L1704) has markdown error messages, but the [English translation](https://github.com/reddit/reddit-i18n/blob/25a519951278fa161cde510bc14222ef8c4faa8c/reddit_i18n/en/LC_MESSAGES/r2.po#L1140-L1145) is still using the old HTML messages, even though the translation site shows the markdown error messages.

It's probably an issue with how we convert what we get from the translation provider to .po files we can use in this repo.

PR #1333, opened 2015-06-10
Media embeds: add vidme to secure services

jbboehr:


eganist:
Can we do some vetting on this before it's automatically included?

I understand that Alexis Ohanian participated in the seed round and all, but the service barely registers in presence. Its mobile apps aren't that well rated and don't have many ratings at all, less than 500,000 downloads on Google Play, and only just got that 3ish million dollar seed round done in April. To my point: it's not battle-tested at all, which means there's no way to judge whether it's actually secure to any measurable degree. Including it in *secure services* would hopefully require a bit more than that.

-Bryant
https://reddit.com/u/eganist/

abenzer:
Bryan, completely understand the concern. We're hoping to get vid.me on this list because we've had many redditors/vidme users report that vidme video embeds on reddit don't appear when reddit is being accessed via HTTPS, unlike youtube and the other sites on the list. We just want to provide the best experience possible for redditors that are watching vidme videos on reddit.

We're still quite new and building our presence, but we are now sub-1000 on Alexa and have significant press visibility. Also 100% transparent and glad to answer any questions you have about our product, team, track record, etc.

In the meantime we'll look into the security auditing services, though admittedly these are kind of high lift for a startup our size. If there are any specific security concerns related to iframe-embedded videos playing on reddit, maybe we could address those first?

If the reddit dev team thinks we're not ready to be on the list, we totally understand!

eganist:
Nah, I can specifically speak to Cure53 (providing service to a project with which I'm familiar), nVisium (myself a former employee), Denim Group (know a few who count themselves as alum), and Matasano (world class) doing excellent work regardless of the size of the company or product. Take my word for it, having a decent AppSec assessment done early on will save you tons of headache down the road before you end up with loads of technical debt, especially since video and gif hosting is your core offering and is still growing. You guys might not have a need for an AppSec *program* yet, but having insight into where your code can get owned and how to mitigate those risks will help you guys more easily be accepted for this sort of embedding with more than just Reddit.

Once you're in a position to offer bug bounties or even just recognition for findings, that's one thing, but until then, consider that if you're included for auto-embedding and your service is compromised, you become a vector for delivering malware to people browsing Reddit at work. 

That's my bigger concern, frankly. Reddit could be extremely secure, but if they add a service that's... well, not... then Reddit is suddenly only as secure as their weakest link. Given how many people browse Reddit at work, an attacker exploiting a third party "secure service" to land malware on workstations is an entirely plausible event. Happens almost literally all the time.

If you get someone to take a crack at your code and show the report and your fixes to anyone at Reddit who would care, I promise you that everyone will be at ease. Not just you guys or Reddit, but your users, too.

spladug:
Hey @jbboehr, thanks a bunch for the patch. We're looking into ways of modernizing that `_SECURE_SERVICES` list right now and hope to have some more info in the near future.  Stay tuned! :)

PR #1335, opened 2015-06-15
Perserve selected search syntax when updating query and provide a link to search in the converted syntax in the δ

Pokechu22:
This makes it easier to search in alternate syntaxes (such as cloudsearch) by doing 2 things: 

1. Preserving the entered search syntax (rather than setting it back to default each search)
2. Making the &delta; (delta) icon have a link to search in the converted syntax.

*This code is untested*.

### DIFF
--- a/r2/r2/controllers/front.py
+++ b/r2/r2/controllers/front.py
@@ -1158,10 +1158,12 @@ class FrontController(RedditController):
         query = q.query
         recent = q.recent
         sort = q.sort
+        syntax = q.syntax
         def wrapper_fn(thing):
             thing.prev_search = query
             thing.recent = recent
             thing.sort = sort
+            thing.syntax = syntax
             w = Wrapped(thing)
             if isinstance(thing, Link):
                 w.render_class = SearchResultLink
--- a/r2/r2/lib/pages/pages.py
+++ b/r2/r2/lib/pages/pages.py
@@ -1450,8 +1450,9 @@ class SearchPage(BoringPage):
             kw['nav_menus'].append(MenuLink(title=_('enable NSFW results'),
                                             url=over18_url))
 
-        self.sr_facets = SubredditFacets(prev_search=prev_search, facets=facets,
-                                         sort=sort, recent=recent)
+        self.sr_facets = SubredditFacets(prev_search=prev_search,
+                                         facets=facets, sort=sort,
+                                         recent=recent, syntax=syntax)
         BoringPage.__init__(self, pagename, robots='noindex', *a, **kw)
 
     def content(self):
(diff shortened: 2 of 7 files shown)


bsimpson63:
Can you give a series of steps to explain how this would be used and how things would change before and after?

Pokechu22:
:haircut: OK, there's two main parts. (this is the right emoji to use, yes?)

&nbsp;

The first change (37c88f4) makes it so that if you manually switch to an alternate search syntax (cloudsearch syntax most of the time), it keeps it that way.  
Right now, you can manually switch to cloudsearch by adding `&syntax=cloudsearch` to the end of the URL.  The reasons you'd want to do so vary, but [one example from the search tips page](https://www.reddit.com/wiki/search#wiki_cloudsearch_syntax) is filtering by specific dates, eg [`1388534400..1420070400`](https://www.reddit.com/search?q=timestamp%3A1388534400..1420070400&syntax=cloudsearch) to search for all posts between 2014 and 2015.  However, if you edit the text in the search box and then click submit, it removes the `&syntax=cloudsearch` from the URL, resulting in a [plaintext search](https://www.reddit.com/search?q=%28and+timestamp%3A1388534400..1420070400+title%3A%27potato%27%29&restrict_sr=&sort=relevance&t=all) (because the new text isn't valid as lucene).  Which means that you'd either have to add `&syntax=cloudsearch` to the end of the URL each time or edit the URL directly, which is a pain.  
What that commit is supposed to do is make it so that the syntax is kept, so that you can actually edit the query in the search box.  It does this by adding a hidden input, which I *think* is the right way of making it.  I'm not a web dev though, so it might be wrong.

&nbsp;

The second commit, 0bc418c, makes it so that the converted syntax revealed by the `δ` is a link to search in that syntax.  
Right now, when you search, you'll see a `δ` in the corner which shows you the converted query.  For example, a search for [`author:pokechu22 subreddit:bugs link`](https://www.reddit.com/search?q=author%3Apokechu22+subreddit%3Abugs+link&restrict_sr=&sort=relevance&t=all) (in the default syntax) is converted to `(and (field author 'pokechu22') (field subreddit 'bugs') (field text 'link'))`.  I believe this is mostly just a debugging thing.  
The change makes it so that you can click that and search in the converted syntax.  Mostly, this won't be needed, but it might be useful in some cases.  To do this, I needed to update the `search_url` method, as right now it doesn't include the syntax.  I also added the syntax for the "search within /r/subreddit" link (but, that probably will never show up, as structured searches within subreddits are disabled).

Pokechu22:
:haircut: Yep, it should have been `thing.syntax`; fixed.  (I also moved the subreddit links preserving the syntax to the first commit, since that seems to make more sense to me).

Pokechu22:
:haircut: (Re comments on fc0a436c): 
 * For `SearchForm`, it's already set on the [constructor in pages.py](https://github.com/Pokechu22/reddit/blob/fdc37e22014ba098e61f763ba1a7b2e148168dd3/r2/r2/lib/pages/pages.py#L3206). 
 * For `SubredditFacets`, [the constructor in pages.py](https://github.com/Pokechu22/reddit/blob/fdc37e22014ba098e61f763ba1a7b2e148168dd3/r2/r2/lib/pages/pages.py#L3258) has been modified to add the syntax as an optional parameter.  
 * For `SearchResultSubreddit`, [the constructor... or whatever it is... in front.py](https://github.com/Pokechu22/reddit/blob/fdc37e22014ba098e61f763ba1a7b2e148168dd3/r2/r2/controllers/front.py#L1161) was edited to include the syntax.

Pokechu22:
Closing this for now.  I may do more with it later, though.

PR #1336, opened 2015-06-16
Remove whitespace from many EOL's

shalecraig:
Started a PR related to something else, and realized that there was trailing whitespace everywhere.

A result of running `find . -not \( -name .svn -prune -o -name .git -prune \) -type f -print0 | xargs -0 sed -i '' -E "s/[[:space:]]*$//"` over the repo.


bsimpson63:
Sorry, but we don't accept pull requests to fix whitespace (and generally only fix whitespace when modifying related code) because they make it difficult to use git blame.

empyrical:
It also has the potential to throw off git's auto merging of pull requests

shalecraig:
:( I suppose this is reasonable

PR #1338, opened 2015-06-16
Add hidden #night-mode checkbox beneath <body>

traverseda:
The goal is to eventually let subreddits create their own night-mode themes. This just adds a (hidden) checkbox beneath the body tag. The checkbox itself doesn't do anything, it just gets used in a selector.

You can style things based on whether or not the checkbox is selected. As an example,

    #night-mode:selected ~ * .usertext-body

would select the usertest-body class only when the night mode checkbox is checked.

The impact is minimal, two lines of front-end code. Down the line, we add a label for that button, and store it's state and default user-preference.

If this goes over well, I'll look into getting a dev enviroment set up and doing those changes.

### DIFF
--- a/r2/r2/public/static/css/reddit.less
+++ b/r2/r2/public/static/css/reddit.less
@@ -11171,3 +11171,5 @@ body:not(.loggedin) .reply-button,
 body:not(.loggedin) .report-button {
     display: none;
 }
+
+#night-mode{display: none;}
--- a/r2/r2/templates/base.html
+++ b/r2/r2/templates/base.html
@@ -60,6 +60,7 @@
   </head>
 
   <body ${classes(*thing.page_classes())}>
+    <input type="checkbox" id='night-mode'>
     ${self.bodyContent()}
     ${self.javascript_bottom()}
   </body>


13steinj:
>PLEASE only use GitHub for confirmed bugs with the code -- NOT for questions or feature requests.


Also, Subreddits already can make their own nightmode themes, and even make fake checkboxes for them, if their designers are actually knowledgable enough about CSS.

bsimpson63:
@13steinj can you give an example of a subreddit that has a nightmode checkbox, or how you could do it with CSS?

13steinj:
@bsimpson63 I've started doing so for /r/agariocss as shown [here](http://i.gyazo.com/92e09d4995700fc78778a24206b280b4.gif), the code is near the end of the stylesheet, however it's at this time incomplete since depending on body classes the user will need different positions.

traverseda:
It reloads? You're using a plugin?

13steinj:
@traverseda Nope. Look at the url. I switch to the `nm` subdomain (night mode), and in the css made a specific style for that. The checkbox works like this:

A link in the sidebar to the nightmode subdomain is selected and placed next to the real checkboxes. Then a fake checkbox is drawn using the `::before` pseudo element. An `::active` and `::focus` is added to make the ckeckmark itself appear.

For the curious: one of the show this subreddit's theme is also a css based checkbox, the other from gold.

@bsimpson63 The basics of it are this: add a links to the sidebar to http://nm.reddit.com/r/subreddit and http://www.reddit.com/r/subreddit. Place them where the other check boxes are supposed to be; and draw a fake checkbox using the pseudo elements listed above. Hide the respective link on the respective domain by using `html:lang(nm)` and whatnot.

traverseda:
Yeah, I caught that eventually. Edited it before you posted. So now you look like a crazy person ;p

13steinj:
@traverseda I always end up looking crazy on the internet. 

But also, I feel as if a subdomain hack works better in this case since it uses less characters. If you didn't know already, your stylesheet can only be 100Kib, i.e. around 100k chars, and moderators over on /r/ModSupport were asking for a higher limit already. `html:lang(nm)` is a lot less than `#night-mode:selected ~ *` in the long run.

bsimpson63:
@traverseda so are you ok with the pure CSS option suggested by @13steinj, or do you want this pull request to be considered?

traverseda:
Think his is a better way to do it for now. Might reopen this if the css limit gets raised, because I think it would work better with tools like RES

PR #1348, opened 2015-06-28
Fix inconsistencies between /by_id/ and /api/info/

### LINKED ISSUE #1347: Inconsistencies between /by_id/ and /api/info/
[GET /by_id/*names*](https://github.com/reddit/reddit/blob/master/r2/r2/controllers/listingcontroller.py#L668) has an internal flag called `ignore_missing` set to `True`, but [GET /api/info/](https://github.com/reddit/reddit/blob/master/r2/r2/controllers/api.py#L182) inherits the default value as `False`. Is there a specific reason why:

- /by_id/ can't be made open to both Comment and Link
- /api/info/ doesn't want `ignore_missing` to be `True`?

If the limit is 100 I can't see how this would be a problem.

Some examples that show what the issue is:

[/by_id/ which ignores bad names](https://www.reddit.com/by_id/t3_39zje0,t3_39zje1,t3_invalid.json)
[/api/info/ which yields nothing instead](https://www.reddit.com/api/info/.json?id=t3_39zje0,t3_39zje1,t3_invalid)

rtheunissen:
Attempt to fix #1347 

The alternative is to allow comments to be queried using `GET /by_id`.

### DIFF
--- a/r2/r2/controllers/api.py
+++ b/r2/r2/controllers/api.py
@@ -179,7 +179,7 @@ class ApiController(RedditController):
     @pagecache_policy(PAGECACHE_POLICY.NEVER)
     @require_oauth2_scope("read")
     @validate(
-        things=VByName('id', multiple=True, limit=100),
+        things=VByName('id', multiple=True, ignore_missing=True, limit=100),
         url=VUrl('url'),
     )
     @api_doc(api_section.links_and_comments, uses_site=True)


rtheunissen:
:moneybag: for :eyeglasses:

ThorArakis:
https://github.com/reddit/reddit/commit/5a0b537cd541d8546e19bcfc5b73fa89c996ad5d

rtheunissen:
:sparkles:

PR #1355, opened 2015-07-01
Add username example to post sharing

allthefoxes:
From: https://www.reddit.com/r/ideasfortheadmins/comments/3br6bk/add_uusername_to_the_example_addresses_when/

### DIFF
--- a/r2/r2/public/static/js/post-sharing.js
+++ b/r2/r2/public/static/js/post-sharing.js
@@ -65,7 +65,7 @@
         '<input class="post-sharing-recipient-input c-form-control" ' +
                'ref="$shareTo" ' +
                'name="recipient" type="text" ' +
-               'placeholder="name@example.com, name@example.com, /u/ExampleUsername">' +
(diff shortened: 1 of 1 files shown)


allthefoxes:
:eyeglasses: @madbook @bsimpson63

allthefoxes:
On a second thought, it seems better to simply change the post sharing to be more clear about the username feature, but this could be a band-aid in the meantime.

tdohz:
We actually deliberately removed the username example since (a) it was very infrequently used and (b) it can be confusing if you're a newer user and not yet acclimated to the /u/username syntax. It just didn't seem like the potential confusion was worth the tiny number of users who used this feature (especially since we still support it as a hidden feature).

> On a second thought, it seems better to simply change the post sharing to be more clear about the username feature

We _may_ add this back in. If we do decide to more explicitly support username sharing, it would be with UI that makes it very clear what's going on, probably as a separate option in the menu.

allthefoxes:
:+1: user->user sharing should definitely still be support, IMO.

Thanks

PR #1361, opened 2015-07-03
This should help you guys out right about now.

chunkysoup:


allthefoxes:
Seriously, can you not?

Also, making changes on master. Tsk Tsk :-1:

@bsimpson63 @madbook can you close?

PR #1363, opened 2015-07-06
Don't send Cache-Control: no-cache, just do private and max-age=0

ErikDubbelboer:
Now that reddit has fully switched to https I have noticed the annoyance that each time I click a link, then click the back button reddit will refresh the page instead of using the cache like a normal website does when you use the back button.

This is because reddit is sending a Cache-Control: no-cache header. A long time ago browsers decided that if an https page sends this it shouldn't use a cached copy on the back button because of some compatibility issues with some bank websites: https://bugzilla.mozilla.org/show_bug.cgi?id=567365

For this reason other big sites like [news.ycombinator.com](https://news.ycombinator.com) use a Cache-Control: private, max-age=0 so proxies don't cache the response, browsers don't cache it on a page refresh, but the browser is still allowed to cache the page for the back button: http://www.w3.org/Protocols/HTTP/Issues/cache-private.html

I think it's good if reddit also does this. This would mean a better experience for people that don't open links in tabs but instead use the back button. It would probably also lower the amount of requests reddit gets by a bit.

### DIFF
--- a/r2/r2/controllers/reddit_base.py
+++ b/r2/r2/controllers/reddit_base.py
@@ -1224,7 +1224,12 @@ class MinimalController(BaseController):
         would_poison = any((k not in CACHEABLE_COOKIES) for k in dirty_cookies)
 
         if c.user_is_loggedin or would_poison:
-            response.headers['Cache-Control'] = 'private, no-cache'
+            # Don't send no-cache, with https browsers will refresh the page when
+            # the back button is used and no-cache is set. Instead use private and
+            # max-age=0 to prevent proxies from caching the result but allowing the
+            # browser to cache it for the back button. When the page is refreshed
+            # in another way the max-age=0 makes the browser send a new request.
+            response.headers['Cache-Control'] = 'private, max-age=0'
             response.headers['Pragma'] = 'no-cache'
 
         # save the result of this page to the pagecache if possible.  we


JordanMilne:
So this is basically what we're going to do, but it's going to be a little more involved than this:

1. I'm iffy about rolling out new caching rules site-wide without knowing how it'll affect downstream proxies, and how many might start improperly caching logged in responses. I wrote some code over the weekend to detect incidences of caches poisoned with logged in responses, and it's in review now. Then we can gradually roll out the new caching headers and gauge how they affect cache poisoning rates, and if the rate of requests hitting the origin server increases.
2. The `Pragma: no-cache` will also trigger the weird back button behaviour, you have to use `Expires: -1` or a similarly old / invalid `Expires` instead (see https://dxr.mozilla.org/mozilla-central/source/netwerk/protocol/http/nsHttpResponseHead.h#49 .)
3. Our load balancers automatically add `Cache-Control: no-cache` headers in a lot of places, and that will need to be changed too

ErikDubbelboer:
Any update on this? The back button is really broken right now. Especially if you collapsed or expanded some comments, click on a link and use the back button to find the whole page reset again.

JordanMilne:
https://www.reddit.com/r/help/comments/3b981x/http_automatically_redirects_to_https/csvratn?context=9

JordanMilne:
Back button should be fixed on Firefox now, the commits will be open-sourced after I untangle everything from the cache poisoning detection :sun_with_face:

ErikDubbelboer:
Yes the new headers fix the problem in Firefox, thanks.

PR #1368, opened 2015-07-20
Add message to be used for when a subreddit is set to private.

justcool393:
IdeasForTheAdmins post: https://www.reddit.com/r/ideasfortheadmins/comments/3dgtar/have_a_separate_this_subreddit_is_private_message/

Username on reddit is same as here. Have tested this on a reddit instance and works fine.

### DIFF
--- a/r2/r2/controllers/api.py
+++ b/r2/r2/controllers/api.py
@@ -2527,6 +2527,7 @@ class ApiController(RedditController):
 
         kw = {k: v for k, v in kw.iteritems() if k in keyword_fields}
 
+        private_message = kw.pop('private_message')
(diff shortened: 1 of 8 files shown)


bsimpson63:
This looks great, but I thought you were just going to reword the description of the existing field to include something like "users will see this if your subreddit is private"?

justcool393:
Oh, I got confused. I can do that and submit as a different pull request if you want.

bsimpson63:
Please do.

This code looks good and correct, but I think for now it's better to not add an additional field.

justcool393:
Now in #1369

PR #1369, opened 2015-07-20
Add that the description will be used in the private message.

justcool393:
This changes on the description wiki page and the settings page.

The wiki page is now:

>The contents of this page appear in the public subreddit description and when the user does not have access to the subreddit

where before it was:

>The contents of this page appear in the public subreddit description

and on the settings page:

>publicly describe your subreddit for all to see. 500 characters max.

to:

>publicly describe your subreddit for all to see. this is shown when a user does not have access. 500 characters max.


bsimpson63:
:fish:

Thanks, I'll get this merged.

bsimpson63:
3b61136426371e8172afdab3d99b0e616ddc0f4f

PR #1370, opened 2015-07-23
Improving on modtools by giving tb usernotes space

creesch:
toolbox usernotes often run out of space for subreddits relying heavily on them. These are also the subreddits that form a positive image about reddit since they are often very well moderated. 
This will allow these subreddits to use toolbox usernotes properly.

### DIFF
--- a/r2/r2/models/wiki.py
+++ b/r2/r2/models/wiki.py
@@ -60,6 +60,7 @@ special_pages = {
     'config/sidebar',
     'config/stylesheet',
     'config/submit_text',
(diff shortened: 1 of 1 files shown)


xiongchiamiov:
Thanks, merged in as 2adfcb3ace0fcec59fdfa4df08c050aace532c94.  `special_pages` isn't actually needed, since the permissions stuff isn't special (as long as you continue to handle setting it on your side).

There may or may not be a lower-level limit these requests are also running into, but if that's the case we still need this change.

PR #1380, opened 2015-08-10
Change "community" on nsfw warning page to "page"

nty-:
"You must be 18+ to view this community" implies that the entire subreddit is nsfw, which isn't always the case when this page is shown.

### DIFF
--- a/r2/r2/templates/over18interstitial.html
+++ b/r2/r2/templates/over18interstitial.html
@@ -35,7 +35,7 @@
 </%def>
 
 <%def name="interstitial_title()">
-  ${_("You must be 18+ to view this community")}
+  ${_("You must be 18+ to view this page")}
 </%def>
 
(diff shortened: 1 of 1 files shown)


nty-:
To elaborate, this page is sometimes shown when you click a NSFW post in a SFW subreddit. Since the subreddit isn't nsfw, this message is misleading in this circumstance.

13steinj:
While I agree; you would have to make a special case. If it's a page in an otherwise SFW sub, then you use `this submission`; if it's a NSFW sub, then you use `this community`

Pokechu22:
How about "`this content`"?  That seems like it would apply to both posts and subreddits.

13steinj:
@Pokechu22 :+1:

umbrae:
Closing this, we're gonna go ahead and make the change on our end.

PR #1381, opened 2015-08-10
Changed notification about asking for upvotes in titles to be less jokey

nty-:
Since reddit takes vote manipulation pretty seriously, saying that doing so violates "intergalactic law" sounds a little too jokey.


bsimpson63:
Thanks for pointing this out.

The rest of the wording in that string is kind of clunky as well so we're going to try to rework the whole thing.

nty-:
Great, thanks!

 I'd like to suggest that you include a mention about reading the subreddit's rules that the user is submitting to.

PR #1384, opened 2015-08-11
Update ban message

captainmeta4:
As per this modtalk discussion: https://www.reddit.com/r/modtalk/comments/3g6786/-/ctv96dt

Per the contribution guidelines, I (and other mods, if you read that thread) feel this change is meaningful enough to warrant a string change.

### DIFF
--- a/r2/r2/lib/system_messages.py
+++ b/r2/r2/lib/system_messages.py
@@ -125,10 +125,10 @@ def send_ban_message(subreddit, mod, user, note=None, days=None):
         message += "\n\n" + blockquote_text(note)
 
     message += "\n\n" + ("you can contact the moderators regarding your ban "
-        "by replying to this message. **warning**: using other accounts to "
-        "circumvent a subreddit ban is considered a violation of reddit's "
-        "[site rules](/rules) and can result in being banned from reddit "
-        "entirely.")
+        "by replying to this message. \n\n"
+        "**warning**: if you use another account to circumvent this subreddit ban, "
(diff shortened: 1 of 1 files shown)


r2/r2/lib/system_messages.py
```diff
@@ -125,10 +125,10 @@ def send_ban_message(subreddit, mod, user, note=None, days=None):
         message += "\n\n" + blockquote_text(note)
 
     message += "\n\n" + ("you can contact the moderators regarding your ban "
-        "by replying to this message. **warning**: using other accounts to "
-        "circumvent a subreddit ban is considered a violation of reddit's "
-        "[site rules](/rules) and can result in being banned from reddit "
-        "entirely.")
+        "by replying to this message. \n\n"
+        "**warning**: if you use another account to circumvent this subreddit ban, "
+        "it would be considered a violation of reddit's [site rules](http://reddit.com/rules) "
```
xiongchiamiov: I don't have any comment on this PR in general, but this link should probably stay `/rules`: that will keep them on the same subdomain and protocol.

captainmeta4: Reverted to `/rules`

bsimpson63:
Can you explain how this wording will be interpreted differently? I don't really understand.

nty-:
It looks like the new wording is supposed to emphasize the hypothetical nature of the warning.

I personally don't like how the change, however, since I think the passive voice weakens the overall message.

bsimpson63:
OK, then I'm going to close this pull request. Sorry.

PR #1385, opened 2015-08-13
More descriptive titles for Twitter status links (suggest title)

plasticantifork:
**Before this change:**
Suggest title = 'Name on Twitter'

**After this change:**
Suggest title = 'Name on Twitter: "This is my status"'


xiongchiamiov:
Hi, and thanks for your contribution!

However, I don't think we really want to do that much special-casing if we can avoid it.  In general, trusting the site to provide a useful `og:title` is a good option, and the user can always change it to whatever else they want.  You might contact Twitter and recommend they provide more useful metadata titles, since that'll affect a wide variety of sites, not just reddit.

PR #1394, opened 2015-08-25
Made it so users cannot friend themselves.

tinfoilboy:
I noticed when I was using RES that I could make myself my own friend, meaning that the backend did no validation that the user was adding itself. I am my own friend, as shown below on my main account.

Demonstration of Self-friend

The way I can reproduce this is by using reddit enhancement suite, hovering over your name, and clicking the friend button. This would basically give a 403 if a request is even made for that.

### DIFF
--- a/r2/r2/controllers/api.py
+++ b/r2/r2/controllers/api.py
@@ -1037,6 +1037,10 @@ class ApiController(RedditController):
         # had better be the current user.
         if type == "friend" and container != c.user:
             abort(403,'forbidden')
+            
+        # users shouldn't be able to friend themselves.
+        if type == "friend" and c.user == friend:
+            abort(403, 'forbidden')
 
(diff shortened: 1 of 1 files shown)


jedberg:
We intentionally made it so that you can friend yourself.  That way when you go to /r/friends, you can see your own posts.

Deimos:
I've seen this come up before, and apparently a decent number of people also friend themselves so that their own posts stand out in comment threads (though maybe it would be better if that just happened naturally anyway).

tinfoilboy:
I didn't know that was intentional. The friend button could be accessible on your own account as well if that's the case.

tinfoilboy:
It would also be a good idea to have the users comments and posts stand out more by default, as it makes more sense.

Deimos:
I'm going to close this pull request, since I think this behavior is something that we want to keep. Thanks for submitting it though, it's definitely a little strange that this is permitted.

PR #1397, opened 2015-09-02
Fix for "retry thumb" bug caused by bad cache data

corylulu:
As seen here: https://www.reddit.com/r/redditdev/comments/3jaoe9/retry_thumb_appears_to_be_bugged/

Embed.ly would occationally cache data before the thumbnail was available for the page AND reddit's internal cache would also cache the meta data without the thumbnail, causing that link to appear without a thumbnail site wide. And when attempting to use the "retry thumb" function, it would simply look at the internally cached data without the thumb and return nothing again. 

So 2 problems were fixed here: 

1. If the media data does not contain a thumbnail, don't cache it internally, fixing the bug that causes the "retry thumb" to fail every time.
2. If thumbnail cannot be found from Embedly data, do a second attempt using the _ThumbnailOnlyScraper to bypass Embedly's cached data. 

This could have also been accomplished by passing a "forced=true" parameter to the EmbedlyScraper, but since it would have involved quite a few more changes, I though the simpler course of action would be to use the _ThumbnailOnlyScraper instead.

### DIFF
--- a/r2/r2/lib/media.py
+++ b/r2/r2/lib/media.py
@@ -312,6 +312,16 @@ def _scrape_media(url, autoplay=False, maxwidth=600, force=False,
             print "%s made a bad secure media obj for url %s" % (scraper, url)
             secure_media_object = None
 
+        # If thumbnail can't be found, attempt again using _ThumbnailOnlyScraper
+        # This should fix bugs that occur when embed.ly caches links before the thumbnail is available
+        if not thumbnail_image:
+            scraper = _ThumbnailOnlyScraper(url)
+            try:
+                thumbnail_image, preview_object, _, _ = (
+                    scraper.scrape())
+            except (HTTPError, URLError) as e:
+                use_cache = False
+
         if thumbnail_image and save_thumbnail:
             thumbnail_size = thumbnail_image.size
             thumbnail_url = upload_media(thumbnail_image)
@@ -320,7 +330,9 @@ def _scrape_media(url, autoplay=False, maxwidth=600, force=False,
                       thumbnail_url, thumbnail_size)
 
     # Store the media in the cache (if requested), possibly extending the ttl
-    use_cache = use_cache and save_thumbnail    # don't cache partial scrape
+    
+    # don't cache partial scrape and don't cache if thumbnail is absent.
+    use_cache = use_cache and save_thumbnail and thumbnail_image 
     if use_cache and media is not ERROR_MEDIA:
         MediaByURL.add(url,
                        media,
--- a/r2/r2/lib/media.py
+++ b/r2/r2/lib/media.py
@@ -313,12 +313,12 @@ def _scrape_media(url, autoplay=False, maxwidth=600, force=False,
             secure_media_object = None
 
         # If thumbnail can't be found, attempt again using _ThumbnailOnlyScraper
-        # This should fix bugs that occur when embed.ly caches links before the thumbnail is available
-        if not thumbnail_image:
+        # This should fix bugs that occur when embed.ly caches links before the 
+        # thumbnail is available
+        if type(scraper) is _EmbedlyScraper and not thumbnail_image:
             scraper = _ThumbnailOnlyScraper(url)
             try:
-                thumbnail_image, preview_object, _, _ = (
-                    scraper.scrape())
+                thumbnail_image, preview_object, _, _ = (scraper.scrape())
             except (HTTPError, URLError) as e:
                 use_cache = False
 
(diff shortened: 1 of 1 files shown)


corylulu:
I setup a test environment to test these changes and it appears to work. Note, however, I think existing broken links that have been cached won't work. But new posts that receive a "retry thumb" will work.

r2/r2/lib/media.py
```diff
@@ -312,6 +312,16 @@ def _scrape_media(url, autoplay=False, maxwidth=600, force=False,
             print "%s made a bad secure media obj for url %s" % (scraper, url)
             secure_media_object = None
 
+        # If thumbnail can't be found, attempt again using _ThumbnailOnlyScraper
+        # This should fix bugs that occur when embed.ly caches links before the thumbnail is available
+        if not thumbnail_image:
```
Deimos: So I think this probably isn't quite the correct approach, since even if the `_ThumbnailOnlyScraper` was chosen initially (for a site embedly doesn't support), this is going to end up just trying again with exactly the same method if it didn't find a thumbnail the first time.

You could probably either try to move this into the "didn't find anything" case inside `_EmbedlyScraper` itself - probably here: https://github.com/reddit/reddit/blob/master/r2/r2/lib/media.py#L682

Or alternatively if you think it makes sense to be out here as a general fallback for *all* scrapers, it should probably just check to make sure it's not repeating, something like `if not thumbnail_image and not isinstance(scraper, _ThumbnailOnlyScraper):`

r2/r2/lib/media.py
```diff
@@ -312,6 +312,16 @@ def _scrape_media(url, autoplay=False, maxwidth=600, force=False,
             print "%s made a bad secure media obj for url %s" % (scraper, url)
             secure_media_object = None
 
+        # If thumbnail can't be found, attempt again using _ThumbnailOnlyScraper
+        # This should fix bugs that occur when embed.ly caches links before the thumbnail is available
```
Deimos: Minor style nit: we try to follow PEP-8-ish style restrictions, which includes making lines no longer than 79 characters, so this comment would need to be wrapped onto another line: https://github.com/reddit/styleguide/blob/master/python/pep-0008.rst#maximum-line-length

r2/r2/lib/media.py
```diff
@@ -312,6 +312,16 @@ def _scrape_media(url, autoplay=False, maxwidth=600, force=False,
...
+        if not thumbnail_image:
+            scraper = _ThumbnailOnlyScraper(url)
+            try:
+                thumbnail_image, preview_object, _, _ = (
```
Deimos: Another minor style nit here: no need to wrap these lines. That was only done elsewhere because the long variable names put it over that line-length limit. So this can just be:

    thumbnail_image, preview_object, _, _ = scraper.scrape()

r2/r2/lib/media.py
```diff
@@ -312,6 +312,16 @@ def _scrape_media(url, autoplay=False, maxwidth=600, force=False,
             print "%s made a bad secure media obj for url %s" % (scraper, url)
             secure_media_object = None
 
+        # If thumbnail can't be found, attempt again using _ThumbnailOnlyScraper
+        # This should fix bugs that occur when embed.ly caches links before the thumbnail is available
+        if not thumbnail_image:
+            scraper = _ThumbnailOnlyScraper(url)
```
xiongchiamiov: This will also recreate the thumbnail scraper if that's what we were already using in the first place, no?

My initial thought on how we should handle this situation is to make [`for_url()`](https://github.com/reddit/reddit/blob/master/r2/r2/lib/media.py#L492-L504) be a generator instead of returning a singular value - so that if one scraper fails, we can fall through to the next appropriate one, whatever that may be, and keep doing that until we succeed at finding something or there aren't any more scrapers.

Deimos:
Looks good, thanks for doing this. Just the minor worry about repeating `_ThumbnailOnlyScraper` and a couple tiny style things.

:nail_care: 

(take a look at https://github.com/reddit/reddit/wiki/The-Code-Salon for the significance of that emoji)

corylulu:
:haircut:

Deimos:
Looks good, just one tiny syntax thing still but I can fix that up myself before I merge it. We don't generally deploy anything on Friday, but I'll most likely push this out on Monday. Thanks again for doing this.

:fish:

corylulu:
Great, thanks. I would be interested in the syntax things you modify for learning purposes. As I said, my python isn't great. So would be nice to know where I make mistakes.

Deimos:
I've merged it here now, just a tiny bit of syntax/rearrangement (there was actually a bug in it currently with `thumbnail_image` not being defined in the outer scope for that `if` statement): https://github.com/reddit/reddit/commit/3db4066f73753546bfb3cdcf8ab06f05443b7636

Thanks again, I'll get you the Open Sorcerer trophy on the site as well.

PR #1398, opened 2015-09-05
Remove the brick from the submit page

13steinj:
On the submit page, there is a brick before "please be mindful of
reddit's few rules and follow good reddiquette".

This commit removes the brick, removed the title attribute from the
notice's div, and updates the link to go directly to reddit's new
content policy.


umbrae:
:eyeglasses: @Deimos

Deimos:
I think we should definitely get rid of this since it doesn't make any sense any more. There's a couple more things that should be done in here though:

1. Remove the change to `.gitignore`, that's not related
2. Actually fully delete the `/r2/r2/public/static/brick-small.png` file, since it's no longer used.

Otherwise this looks good though, just those two minor things. :nail_care:

13steinj:
@Deimos sorry it took so long; the Windows TH2 update gave me troubles with my VM. Should I squash the commits? :haircut:

kemitche:
Goodbye brick :(

spladug:
:(

kemitche:
I will mourn the brick but I'm not here to block its removal! Insider jokes should be cycled out after a while.

Deimos:
Looks good, thanks. :fish:

Deimos:
Merged and deployed: https://github.com/reddit/reddit/commit/de7ced1a72cca6b031bde554528ed72cec1c9b98

PR #1399, opened 2015-09-07
Prevent duplicate stickies

TheEnigmaBlade:
Previously a link was able to be stickied in both sticky slots simultaneously, resulting in a link being displayed twice in a subreddit's listing. This would have some interesting results, as seen in /r/ffxiv:

### DIFF
--- a/r2/r2/models/subreddit.py
+++ b/r2/r2/models/subreddit.py
@@ -1303,6 +1303,10 @@ class Subreddit(Thing, Printable, BaseSite):
             #   saved properly unless we assign a new list to the attr
             sticky_fullnames = self.sticky_fullnames[:]
 
+            # if the link is already stickied, remove it to prevent duplicates
+            if link._fullname in sticky_fullnames:
+                sticky_fullnames.remove(link._fullname)
+
             # if a particular slot was specified and is in use, replace it
             if num and num <= len(sticky_fullnames):
                 unstickied_fullnames.append(sticky_fullnames[num-1])


Pokechu22:
Second example: [/r/pokechu22](https://www.reddit.com/r/pokechu22/).  It's also fun that when you vote on one, it simultaneously changes the arrow on the other.

[/r/bugs post](https://www.reddit.com/r/ModSupport/comments/3jwy99/bug_duplicate_stickies/).  

And to manually reproduce the bug: 

1. Make sure your subreddit has no stickies at the moment.
2. Find a post to sticky.
3. Open that post twice.
4. In the first tab, sticky the post.
5. In the second tab, sticky the post.
6. Go back to the subreddit.  The same post will be stickied twice.  

(Note that that doesn't make sense for /u/automodetor, though)

Deimos:
Hmm, I think this probably isn't quite the right behavior, but I guess it depends exactly what the right behavior is. If we have Link A and Link B stickied (in that order, so Link B is on the bottom), and someone tries to re-sticky Link A, what should happen?

With this code, you'd basically end up swapping the two positions - Link B would move to the top, and Link A would become the bottom sticky. My inclination is that that probably shouldn't happen, that things should probably just stay exactly the same.

I'll ask some other opinions, do you think the swapping is the correct behavior?

:nail_care:

TheEnigmaBlade:
I guess it depends on how you think about the concept of sticky positions:

* If you consider the two sticky slots as two elements in a list, this would be the correct behavior.

* If you consider the two slots to be two concrete locations accessed using absolute indices, this would be the wrong behavior.

The first behavior is derived from the current queue-style behavior: new and replacement stickies are always added to the end (second sticky slot), and removing a sticky shifts more recent stickies upwards. This way the behavior introduced in this change could be considered more of a "re-sticky" than a swap. If a link is attempted to be stickied twice, it's actually stickied twice (once normally and once a re-sticky).

The second behavior is (and I'm having trouble organizing it into words) based on a thing model where things are placed into physical slots. In other words, links are assigned to specific sticky locations rather than acquiring a location as a side effect of an append operation. Swapping positions through a remove and append doesn't make sense in this case since the sticky was assigned a location; a second sticky operation should do nothing.

If mods were allowed to specify the locations of stickies through the default UI rather than all operations being performed through a series of toggles, I would be more inclined to side with the second behavior. But based on the mental model I developed before looking at the code -- where the toggles add/remove rather than set/remove -- the first behavior makes much more sense.

So yes, I believe the swapping behavior is correct.

If I understand this weird emoji thing correctly... :haircut:

Deimos:
I think the main problem I have with this is related to the fact that the UI doesn't allow you to do it unless you do something unusual like @Pokechu22 described above, where you open two copies of the link before stickying it in the first place. So while the swapping behavior makes sense with the "queue model" in theory, in reality I think this is really only ever going to happen by accident.

It will probably only happen in cases like two moderators trying to sticky the same post almost simultaneously, a bot/script repeatedly trying to sticky the same post even though it's already there, and so on. So in pretty much all the cases that will actually come up, I think the expected result would be that nothing happens, not that the order reverses.

Also, the behavior of this current code will definitely be wrong in a case where the user is specifying the top slot (currently only possible through the API) and the same post is already there, with another one in the bottom slot. If Link A is the top sticky and Link B is the bottom, and you do a request to sticky A in the top slot, this could would give:

1. A is removed
2. B moves up into top slot
3. New A replaces B in the top slot

So it would end up with only A stickied, which I definitely don't think is the expected result. This is definitely something that could happen with a script of some sort if it tries to put something into the top sticky slot without recognizing that it's already there.

My inclination is that the result should be that nothing happens if you try to sticky an already-stickied link, or probably even better have it return an error that it's already stickied.

(and yes, you did the emoji thing right) :nail_care:

Deimos:
I'm going to do up a pull req internally myself for the other approach (confirmed that @TheEnigmaBlade was okay with me taking it over).

PR #1401, opened 2015-09-07
Fix vanishing data-fullname attribute from modmail messages.

TheEnigmaBlade:
The addition of subreddit colors introduced the `thing_data_attributes` def to the `message` template, which, according to my quick research about inheritance in mako, overrides the def of the same name in the parent `printable` template. This caused the `data-fullname` attribute to suddenly vanish and set many toolbox modmail features on fire.

As in the `thing_css_rowclass` def right above, the parent's `thing_data_attributes` should be executed to ensure data attributes like `data-fullname` are added.

### DIFF
--- a/r2/r2/templates/message.html
+++ b/r2/r2/templates/message.html
@@ -61,6 +61,7 @@
 </%def>
 
 <%def name="thing_data_attributes(what)">
+  ${parent.thing_data_attributes(what)}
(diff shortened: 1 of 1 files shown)


florenceyeun:
:fish: Thanks for the fix.

florenceyeun:
a5e489d4169fd9842b76fbd50159134d800ee6f6

PR #1403, opened 2015-09-11
Fix automod config perm issue on sidecontent box

13steinj:
Resolves #1402

PR #1404, opened 2015-09-11
Fix for incorrect 'config' permissions related to automoderator config

### LINKED ISSUE #1402: Moderation Tools shows "automoderator config" with insufficient permissions
**Issue**: A moderator with 'wiki' permissions and not 'config' permissions can see the "automoderator config" option within moderation tools. This should not be present as 'config' permissions are required to edit AutoMod. Furthermore clicking that label sends the insufficient moderator to a forbidden page. 

**How to Reproduce**: Grant a moderator just 'Wiki' permissions on a subreddit with AutoMod already set up. Click the button and receive "forbidden" error.

**Expected Result**: For just 'wiki' perms "automoderator config" links to the non editable wiki page (wiki/config) and for anyone with 'config' perms the button links to the wiki/edit/config page.

**Fix in #1405**

english06:
Resolves #1402 by granting mods with 'config' permissions the ability to see "automod config" in the moderation tools sidebar and link to the editor. A more robust solution than #1403 as it retains 'wiki' mods access to that button and instead links them to a viewable page.

### DIFF
--- a/r2/r2/lib/pages/pages.py
+++ b/r2/r2/lib/pages/pages.py
@@ -501,19 +501,7 @@ class Reddit(Templated):
             if is_moderator_with_perms('flair'):
                 buttons.append(NamedButton("flair", css_class="reddit-flair"))
 
-        if is_single_subreddit and is_moderator_with_perms('wiki'):
-            # append automod button if they have an AutoMod configuration
-            try:
-                WikiPage.get(c.site, "config/automoderator")
-                buttons.append(NamedButton(
-                    "automod",
-                    dest="../wiki/config/automoderator",
-                    css_class="reddit-automod",
-                ))
-            except tdb_cassandra.NotFound:
-                pass
-
-        if is_single_subreddit and is_moderator_with_perms('config'):
+        if is_single_subreddit and is_moderator_with_perms('wiki') or is_moderator_with_perms('config'):
             # append automod button if they have an AutoMod configuration
             try:
                 WikiPage.get(c.site, "config/automoderator")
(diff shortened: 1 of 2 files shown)


13steinj:
@english06 This is not correct. You added code to add a second button if you have config; but otherwise it will show a button to the regular page no matter what. I've fixed this in #1403 regardless.

english06:
@13steinj You were right. I added a check in the 'wiki' mod paragraph to not fire for 'config' perm mods. The problem with #1403 is that 'wiki' mods simply get no button at all. This request accounts for them and allows them to have a quick link to view the automod config to which they naturally already have access (just not to edit).

13steinj:
Wiki mods can not view the config page AFAIK; let me check.

english06:
'Wiki' perms allow them to view the page. They are just not allowed to edit it. That throws back an error:

>forbidden (reddit.com)

>you are not allowed to do that
— may_not_revise.

13steinj:
@english06 you're right, but there is no reason for a wiki mod person to have a button. There is also no reason to have two different possible buttons with the same name and different functionality.


Also, because your PR was made before @Deimos changed the button to link directly to the edit page; if your change is merged from how I know github works it would make both buttons link to the edit page.

english06:
The buttons will be the same as they have always been and simply send them to either the view page or edit page depending on their permission level. This is a seamless approach. 

As for the recent change how should I go about accounting for that in 510?

english06:
I am closing this to gather all commits into single action and account for the update to 510

13steinj:
@english06 you could have just rebased to master; but even if you do that, the button for the wiki mod makes no sense whatsoever.

PR #1405, opened 2015-09-12
Fix for incorrect 'config' permissions related to automoderator config 

### LINKED ISSUE #1402: Moderation Tools shows "automoderator config" with insufficient permissions

english06:
Grants mods with 'config' permissions the ability to see "automoderator config" in the moderation tools sidebar and link to the editor. Previously this button only showed for 'wiki' mods which did not match with prior documentation that allows this action for 'config' mods. Furthermore this resolves #1402 by linking wiki mods to the viewable URL instead of the edit one to which they do not have permission.

A more robust solution than #1403 as it seamlessly retains 'wiki' mods access to that "automoderator config" button and instead links them to a viewable page instead of just simply giving the access to 'config' mods and leaving 'wiki' mods with a shortcut.

### DIFF
--- a/r2/r2/lib/pages/pages.py
+++ b/r2/r2/lib/pages/pages.py
@@ -501,8 +501,20 @@ class Reddit(Templated):
             if is_moderator_with_perms('flair'):
                 buttons.append(NamedButton("flair", css_class="reddit-flair"))
 
-        if is_single_subreddit and is_moderator_with_perms('wiki'):
-            # append automod button if they have an AutoMod configuration
+        if is_single_subreddit and is_moderator_with_perms('wiki') and not is_moderator_with_perms(‘config’):
+            # append automod button if they have an AutoMod configuration for 'wiki' perms
+            try:
+                WikiPage.get(c.site, "config/automoderator")
+                buttons.append(NamedButton(
+                    "automod",
+                    dest="../wiki/config/automoderator",
+                    css_class="reddit-automod",
+                ))
+            except tdb_cassandra.NotFound:
+                pass
+
+        if is_single_subreddit and is_moderator_with_perms('config'):
+            # append automod button if they have an AutoMod configuration for 'config' perms
             try:
                 WikiPage.get(c.site, "config/automoderator")
                 buttons.append(NamedButton(


13steinj:
Like I said in #1404 adding the wiki mods the shortcut is both inconsistent and makes no sense. Wiki mods do not have the shortcut for anything else; so why should tehy have it for the automod config? They also would have no use for it.

english06:
I don't understand why you are so adamant against this. AutoMod is a *huge* part of moderating and having a shortcut to view AutoMod's config is extremely helpful. This commit doesn't change or add anything new, it just fixes a bug.

13steinj:
I'm "adamant" about this because 1, inconsistencies are never a good thing, and 2, even if it is a huge part of moderating (which it is), a wiki mod gains absolutely nothing from having a quick and easy shortcut, except that shortcut, which is not useful to him or her, since that page is not useful to him/her ***whatsoever***.

english06:
'Wiki' mods already have access to this config, this doesn't change that. As a mod of a 3.1+ million subscriber subreddit where I have 'wiki' perms without 'config' let me say the shortcut is extremely valuable. I would not be able to moderate without viewing that config a couple of times per day to see what domains we screen, or other settings we have dialed in. Right now to view this config I have to use a

13steinj:
...the comment you made ends in the middle of a sentence.

13steinj:
Update: it no longer ends in the middle of a sentence; continue your peace after reading mine.

english06:
Because anyone can view sidebar or stylesheet of any subreddit. You don't have to be a mod for that. Look I know you don't understand and I can't explain it to you. Your largest sub is 17,000 and that's fine. This is not a big deal and I am moving on. If the dev team wants this commit then wonderful. All I am doing here is fixing a bug. /endreplies

13steinj:
"Your largest subreddit is 17,000 and that's fine."

Firstly, that's not the point. You're acting like that gives you some kind of privilege. You have been a mod of that sub (which I will keep the name of private) for 12 damn days. Less than two damn weeks. Don't act high and mighty. Even if you were a mod of it for much much longer; it still gives you no privilege. You also are considering that I have no alts; and am not a mod of a larger sub. For the sake of telling, no, I do not mod a larger sub; but this matter is of principle (which I also used on you).

It's not that I don't understand; it's that you don't understand the matter of the fact that it's highly inconsistent, and has no actual use unless you can prove a use; which you have just now given up in doing.

Yes, it isn't a big deal, but it is still a matter of consistency and principle. You are not "fixing a bug". You are correcting a bug (which if I may remind you, I'd bet if I had not gone and done it for you first and notify you, you would either take forever to find it or wouldn't find it at all); while also adding some absolutely useless functionality of your own; that is also inconsistent with previous matter.

Additional functionality should not really be in this repo according to the README; but technically no one is stopping it. Even so, that's for /r/ideasfortheadmins / a message to /r/reddit.com. 

This repo should be for reporting confirmed issues, and or fixing them.

"/endreplies"

What, is this some kind of grade schooler chatroom?

umbrae:
Thanks for the contribution, I think we're going to pass on this for now.

Also as an aside (to both of you), arguing isn't tolerated in pull requests here, as it can be a chilling effect to other folks who may want to contribute but fear the discussion around it. Feel free to flame each other via reddit PM next time.

PR #1408, opened 2015-09-13
Correct md_sender in "gold gift" messages.

13steinj:
Previously, due to how Snudown works, if a username had two underscores, it would be rendered as if anything within the underlines would be italicized, and the underlines dropped. This makes it so that the italicizing effect caused by underlines no longer makes this render incorrectly; at the minimal cost of making the sender's name in `/u/` format, (which in reality, affects absolutely nothing in terms of meaning of the message/sender)

[BUG REPORT](https://www.reddit.com/r/bugs/comments/3krzm8/username_is_displayed_incorrectly_in_guild/)

### DIFF
--- a/r2/r2/controllers/ipn.py
+++ b/r2/r2/controllers/ipn.py
@@ -279,7 +279,7 @@ def send_gift(buyer, recipient, months, days, signed, giftmessage,
 
     if signed:
         sender = buyer.name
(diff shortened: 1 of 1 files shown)


MelissaCole:
Looks good. Thanks!

MelissaCole:
https://github.com/reddit/reddit/commit/7874382d1c9df1e197661a2547b9c913ef93ff54

PR #1410, opened 2015-09-15
Doesn't use cache metadata if thumbnail missing

corylulu:
Still instances where data is being cached with invalid thumbnail. Couldn't pinpoint where it might be getting through, so added additional checks when grabbing the cached data to ensure it's valid. 

I could also do this in the '/r2/r2/models/media_cache.py' file, I believe, but didn't know if that would be better.

### DIFF
--- a/r2/r2/lib/media.py
+++ b/r2/r2/lib/media.py
@@ -284,8 +284,8 @@ def _scrape_media(url, autoplay=False, maxwidth=600, force=False,
         if mediaByURL:
             media = mediaByURL.media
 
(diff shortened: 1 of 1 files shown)


umbrae:
:eyeglasses: @Deimos

Deimos:
This seems reasonable, sorry for missing this earlier. :fish:

Deimos:
Thanks, merged and deployed: https://github.com/reddit/reddit/commit/e152ff6ad9260d243a8d4e28d9512fb94cd73348

PR #1411, opened 2015-09-16
Change so mods can't block their own subreddits

13steinj:
Previously, a mod would be able to block their own subreddit in their modmail, if they saw a ban message / original mod invite message.

This changes it so the block subreddit buttons would only show the block subreddit button if the user is not a mod of the subreddit.

### DIFF
--- a/r2/r2/controllers/api.py
+++ b/r2/r2/controllers/api.py
@@ -1709,7 +1709,9 @@ class ApiController(RedditController):
             sr = None
 
         if getattr(thing, "from_sr", False) and sr:
-            BlockedSubredditsByAccount.block(c.user, sr)
+            # Users may only block a subreddit they don't mod
+            if not (sr.is_moderator(c.user) or c.user_is_admin):
+                BlockedSubredditsByAccount.block(c.user, sr)
             return
 
         # Users may only block someone who has
--- a/r2/r2/templates/printablebuttons.html
+++ b/r2/r2/templates/printablebuttons.html
@@ -453,17 +453,21 @@
     %if thing.can_block:
       ${self.banbuttons()}
         %if (not thing.was_comment or thing.thing.is_mention) and thing.thing.author_id != c.user._id and thing.thing.author_id not in c.user.enemies:
-          <li>
-            %if getattr(thing.thing, "from_sr", False):
-              %if getattr(thing.thing, "sr_blocked", False):
-                ${ynbutton(_("unblock subreddit"), _("unblocked"), "unblock_subreddit")}
-              %else:
-                ${ynbutton(_("block subreddit"), _("blocked"), "block", "hide_thing")}
-              %endif
-            %else:
-              ${ynbutton(_("block user"), _("blocked"), "block", "hide_thing")}
+          %if getattr(thing.thing, "from_sr", False):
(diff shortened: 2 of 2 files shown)


r2/r2/templates/printablebuttons.html
```diff
@@ -451,17 +451,21 @@
...
-            %if getattr(thing.thing, "from_sr", False):
-              %if getattr(thing.thing, "sr_blocked", False):
-                ${ynbutton(_("unblock subreddit"), _("unblocked"), "unblock_subreddit")}
-              %else:
-                ${ynbutton(_("block subreddit"), _("blocked"), "block", "hide_thing")}
-              %endif
-            %else:
-              ${ynbutton(_("block user"), _("blocked"), "block", "hide_thing")}
+          %if getattr(thing.thing, "from_sr", False):
+            %if not (thing.thing.subreddit.is_moderator(c.user) or c.user_is_admin):
```
MelissaCole: You can use `thing.user_is_moderator` - https://github.com/reddit/reddit/blob/master/r2/r2/models/link.py#L2031

MelissaCole:
I'm trying to think about if there's any reason a mod would want to block mail from their subreddit. I just ran a query and 11 users have blocked one of the subreddits that they mod, so it's probably a small enough number to merit not including that button.

To completely prevent the blocking of subreddits they mod, you'll have to check if the user is a moderator [here](https://github.com/reddit/reddit/blob/master/r2/r2/controllers/api.py#L1711) as well

13steinj:
@MelissaCole forgot about the API! I'll add that in and change the %if check on printablebuttons.html as to your suggestion. Those 11 users are probably the people that tested it after the /r/ModSupport thread was made.

13steinj:
@MelissaCole done.

r2/r2/controllers/api.py
```diff
@@ -1709,7 +1709,9 @@ def POST_block(self, thing):
             sr = None
 
         if getattr(thing, "from_sr", False) and sr:
-            BlockedSubredditsByAccount.block(c.user, sr)
+            # Users may only block a subreddit they don't mod
+            if not (thing.user_is_moderator or c.user_is_admin):
```
MelissaCole: The wrapped properties won't be available here (`user_is_moderator`) since it's a Thing being returned in the validator, so it will have to be `sr.is_moderator(c.user)`

13steinj:
@MelissaCole Okay, now done. I hope.

MelissaCole:
:fish: Looks good. Thanks! Will push it out tomorrow

13steinj:
:+1:

MelissaCole:
https://github.com/reddit/reddit/commit/787fcdc8d0e3bd1dd5ac8dfa8186df2bdde41c5a

Also ran script to remove blocks that mods made of their own subreddits since now they can't unblock it.
```
for x in list(BlockedSubredditsByAccount._cf.get_range()):
    for key in Subreddit._byID36(x[1].keys()):
        if Subreddit._byID36(key).is_moderator(Account._byID36(x[0])):
            print getattr(Account._byID36(x[0]), 'name', 'no name'), Subreddit._byID36(key)
            BlockedSubredditsByAccount.unblock(Account._byID36(x[0]), Subreddit._byID36(key))
```

PR #1414, opened 2015-09-19
Add an API endpoint to get the subreddits moderated by a user.

Pokechu22:
This has [been](https://www.reddit.com/r/redditdev/comments/35v5h5/is_there_an_api_call_for_getting_list_of_a_users/) [requested](https://www.reddit.com/r/redditdev/comments/2c64ed/praw_can_i_get_the_list_of_subrredits_moderated/) [several](https://www.reddit.com/r/redditdev/comments/3gae4j/how_can_i_get_a_list_of_subreddits_a_user/?) [times](https://www.reddit.com/r/redditdev/comments/3jx0ik/request_can_we_get_an_endpoint_for_a_users_mod) in /r/redditdev.  It adds an endpoint at `/user/<username>/moderated`.

### DIFF
--- a/r2/r2/controllers/listingcontroller.py
+++ b/r2/r2/controllers/listingcontroller.py
@@ -30,7 +30,7 @@ from r2.models import *
 from r2.models.query_cache import CachedQuery, MergedCachedQuery
 from r2.config.extensions import is_api
 from r2.lib.filters import _force_unicode
-from r2.lib.jsontemplates import get_usertrophies
+from r2.lib.jsontemplates import get_usertrophies, TrimmedSubredditJsonTemplate
 from r2.lib.pages import *
 from r2.lib.pages.things import wrap_links
 from r2.lib.menus import TimeMenu, SortMenu, RecSortMenu, ProfileSortMenu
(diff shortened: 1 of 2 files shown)


13steinj:
<3

I can't really tell, but hopefully this is only the *public* modlist, correct?

Pokechu22:
> I can't really tell, but hopefully this is only the *public* modlist, correct?

Yep.

    mod_srs = [sr for sr in all_mod_srs if sr.can_view_in_modlist(c.user)] 

Same logic that's used to display the actual modlist.

13steinj:
:+1:

r2/r2/controllers/listingcontroller.py
```diff
@@ -977,6 +977,29 @@ def GET_trophies(self, user):
...
+        all_mod_srs = Subreddit._byID(mod_sr_ids, data=True,
+                                      return_dict=False, stale=True)
+        mod_srs = [sr for sr in all_mod_srs if sr.can_view_in_modlist(c.user)]
+
+        if not mod_srs:
+            return self.api_wrapper([])
+
+        templ = TrimmedSubredditJsonTemplate()
+        resp = [templ.render(sr).finalize() for sr in mod_srs]
+
```
13steinj: @Pokechu22 Forgive my ignorance if this is incorrect as I haven't tested your branch on my local install; but I was playing around with templates just now; and the values for the dictionary keys came out in unicode (`u'text here' format`); so you may have that problem as well (unless `api_wrapper()` takes care of that?)

Pokechu22: I haven't tested it in a while, but I'm guessing it does.  Based off of the fact that the trophies code is basically the same, and gives the right output ([example](https://www.reddit.com/user/pokechu22/trophies.json)).

13steinj: @Pokechu22 Yep, it does, and it was the missing key to making get_sidebar to work as well.

umbrae:
We're going to talk about this internally.

13steinj:
Heh, I didn't get a notification.

Now I know my say makes no difference, but I still want to put it out there: we already can view a user's moderated subreddits in the sidebar, and we can view their public multireddits in the same fashion, as well as view their public multireddits in an API response. This change would be very useful and if anything makes it consistent with the above.

umbrae:
Thanks for the contribution - right now we're looking into possibly doing this in a lighter weight way. It would be handy but subreddits endpoints are a bit heavy on load for us. We haven't quite figured that out right now though, so we're going to close this for now.

bb010g:
Has there been any progress on doing this in a lighter way?

13steinj:
@bb010g call me cynical but "lighter way" doesn't make sense-- it's the same exact process loading these subreddits on a user page.

PR #1415, opened 2015-09-21
Add CSS class indicating deleted vs removed status

alexanderpas:
This allows CSS to differentiate between posts that are deleted by the user, and posts that are removed by a moderator or administrator

### DIFF
--- a/r2/r2/models/link.py
+++ b/r2/r2/models/link.py
@@ -1436,8 +1436,10 @@ class Comment(Thing, Printable):
                     # If removed by an admin or moderator, distinguish that
                     # from being deleted by the user.
                     if item._spam:
+                        extra_css += " removed"
                         item.body = '[removed]'
                     else:
+                        extra_css += " deleted"
                         item.body = '[deleted]'
 
             if focal_comment == item._id36:


13steinj:
Isn't this already done via a lack of the "spam" class vs no extra class at all? And this is more of a design thing, than a user notice thing.

And this would differentiate between an admin takedown and the others, but that's already differentiated regardless, in a way.

Umdlye:
@13steinj , the spam class is only for moderators. This change will allow moderators to style user-deleted posts and mod-removed posts separately.

13steinj:
@Umdlye K then I guess? Not much of a point; in fact it opens it up to more abuse in a way. People can already see the difference via the difference in text.

minecast:
@13steinj It's for moderators to use. This would allow mods to do different things with removed or deleted posts, such as hiding them, or styling them differently, etc. I don't see how this can possibly be open to abuse. I'm in agreement with @Umdlye.

13steinj:
@minecast 

> Hiding them

That's abuse in and of itself. I've no idea what side of the "transparency spectrum" you are on now

minecast:
@13steinj I didn't have any examples at the time when I said that. It would make the comments cleaner if a mod nukes a thread. Plus the user can just disable the CSS. If this gave mods the ability to hide removed comments, then they can do it now anyway. It's just more helpful.

13steinj:
> make comments cleaner

Yup. Also lacks transparency.

> plus a user can just disable css

Not easily.

> they can do it now anyway

Also, not easily.

minecast:
@13steinj "Also, not easily."
This wouldn't change the difficulty.

"Not easily."
There is a checkbox you uncheck. Users can uncheck it.

"lacks transparency"
If the intention of the mods was to hide the fact they remove comments, they could have done it already. This is just so they can style it differently and possibly reduce confusion.

Pokechu22:
@minecast

> This wouldn't change the difficulty.

I can see what @13steinj means -- it changes it from having to manually specify comment IDs to being able to simply do `.removed { display: none; }`.  But if they were that concerned about hiding comments, they would just do `.grayed { display: none; }` and not worry about getting both deleted and removed comments.

> There is a checkbox you uncheck. Users can uncheck it.

That's only available for gold users, on a per-subreddit basis at least.  The other checkbox is in your preferences, and affects all subreddits, which most users probably wouldn't want.  (There's also the RES checkbox, but I'm going to ignore the existence of that for now)

> If the intention of the mods was to hide the fact they remove comments, they could have done it already. This is just so they can style it differently and possibly reduce confusion

Agree with this.  I don't see how allowing styling reduces transparency.

13steinj:
> This wouldn't change the difficulty.


Yes it would. 

```
.removed {display:none;}

v.s.

<special selectors based on fullnames>  {display:none;}
```

If the first isn't easier, I don't know what is/isn't.

>There is a checkbox you uncheck. Users can uncheck it.

That's RES / Reddit Gold only. Other than that the user would have to go through /prefs, which disables styles completely.

And what if I want the styles?

>If the intention of the mods was to hide the fact they remove comments, they could have done it already. This is just so they can style it differently and possibly reduce confusion.

What. With that argument they can also currently style it differently  as is, however difficult-- and that's besides the point. Styling, I couldn't care less. But this makes it easy do just do a `{display:none;}` block.

---

Putting all that aside, you do realize you've reawakened a month old thread, right? Just a weird thing to do.

umbrae:
We've got something like this going on in-house, so we're going to close this one.

Additionally, we don't allow argument in pull requests, as we want to be inviting for users to consider adding to reddit. Please take it easy.

Issue #1240, opened 2015-01-16
adding a meta tag to open in AlienBlue on iOS


Haroenv:
the meta tag should be in the html and promoted like this:

    <meta name="apple-itunes-app" content="app-id=myAppStoreID, affiliate-data=myAffiliateData, app-argument=myURL">

[apple document](https://developer.apple.com/library/ios/documentation/AppleApplications/Reference/SafariWebContent/PromotingAppswithAppBanners/PromotingAppswithAppBanners.html)

jvbreen1:
This issue can probably be closed.

PR #1416, opened 2015-09-27
Add meta tag for AlienBlue in iOS

the-hobbes:
As per issue #1240, add a meta tag to support promotion of AlienBlue via Smart App Banners for iOS.

### DIFF
--- a/r2/r2/templates/base.html
+++ b/r2/r2/templates/base.html
@@ -34,6 +34,7 @@
     <meta name="keywords" content="${self.keywords()}" />
     <meta name="description" content="${getattr(thing, 'short_description', None) or g.short_description}" />
     <meta name="referrer" content="${c.referrer_policy}">
+    <meta name="apple-itunes-app" content="app-id=923187241">
     <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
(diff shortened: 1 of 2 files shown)


english06:
Is there a way to add the "Navigational Context" as well?

https://developer.apple.com/library/prerelease/ios/documentation/AppleApplications/Reference/SafariWebContent/PromotingAppswithAppBanners/PromotingAppswithAppBanners.html

the-hobbes:
Sure- just so I make sure I'm understanding correctly, we'd want to pass something like "https://www.reddit.com/#page=2" as the app-argument to the app so that the user can open the app at the same position she was in at the time of the smart banner click?

english06:
Well even beyond that though. To submission pages, comment pages, messages, modmail. I think those 4 would cover most traffic. Going to specific content may be tough depending on Alien Blue design, but ideally just the general pages shouldn't.

the-hobbes:
I'm passing the entire request url, with the intent that the iOS app will handle the navigation context routing.

umbrae:
Thanks for the contribution, unfortunately we're not ready for this yet. We'll definitely be thinking about something like this at some point though.

PR #1418, opened 2015-10-02
Don't make things with NSFW in the title forced

13steinj:
Resolves [this](https://www.reddit.com/r/ModSupport/comments/3n46i3/request_allow_mods_to_unnsfw_posts_regardless_of/). This allows for posts that have nsfw in the title to determine the over_18 attribute, *once*, at creation, instead of forcing posts to be nsfw via the over_18 attribute "or" the fact that it has the phrase "nsfw" in some form. Because of the fact that it only does this at creation, the over_18 attribute can be changed successfully, and the previous or statement takes no effect (and as such is removed).

Edit: Just realized the logic is ever so slightly off, so don't merge yet. As it is, any currently made nsfw post that doesn't have nsfw in the title may be auto unnsfwd, due to me adding the dummy check and that check is added, and then when the check needs to run the first time it will determine over_18 only based on the property. A check for `if not self.over_18` before setting over_18 should correct it though, or you can do some server side magic to get all current nsfw submissions and set their over_18 value to True after implementing.

### DIFF
--- a/r2/r2/models/link.py
+++ b/r2/r2/models/link.py
@@ -217,10 +217,8 @@ class Link(Thing, Printable):
             url = content
             selftext = cls._defaults["selftext"]
 
-        nsfwcheck = re.compile(r"\bnsf[wl]\b", re.I)
         over_18 = False
-
-        if bool(nsfwcheck.search(title)):
+        if cls._nsfw.search(title):
             over_18 = True
 
         l = cls(
@@ -236,7 +234,7 @@ class Link(Thing, Printable):
             ip=ip,
             comment_tree_version=cls._choose_comment_tree_version(),
             is_self=is_self,
-            over_18=over_18
+            over_18=over_18,
         )
 
         l._commit()
--- a/r2/r2/models/link.py
+++ b/r2/r2/models/link.py
@@ -217,6 +217,12 @@ class Link(Thing, Printable):
             url = content
             selftext = cls._defaults["selftext"]
 
+        nsfwcheck = re.compile(r"\bnsf[wl]\b", re.I)
+        over_18 = False
+
+        if bool(nsfwcheck.search(title)):
+            over_18 = True
+
         l = cls(
             _ups=1,
             title=title,
@@ -230,6 +236,7 @@ class Link(Thing, Printable):
             ip=ip,
             comment_tree_version=cls._choose_comment_tree_version(),
             is_self=is_self,
+            over_18=over_18
         )
 
         l._commit()


Deimos:
:eyeglasses: @Deimos

13steinj:
Okay; that should fix it from what I can tell. Might be safe to have some kind of backup either way.

Deimos:
Hmm, I think this is over-complicating it quite a bit. Would it work to just do the regex check at the time of submission (probably in the `Link._submit` function), and just set `over_18` on the link as it's being created there? It seems like that would cover it to me, and then we could drop the `nsfw_str` attribute entirely. Is there something about doing it that way that wouldn't work? :nail_care:

13steinj:
@Deimos 

I'm never going to understand emojis.

While under normal circumstances that definitely would work; it wouldn't in the case of the actual reddit.com since there already were posts with the words "NSFW" in their title.


If the check is moved to the `_submit` function; then wouldn't every single post that is currently marked NSFW because it has "NSFW" in the title become un-nsfwd?

Deimos:
The emoji is something that we use to manage our code review process: https://github.com/reddit/reddit/wiki/The-Code-Salon So my `:nail_care:` is basically me saying "I'm done looking at this right now, waiting for responses/updates from the author", and using that emoji updates various internal stuff related to open pull requests (for example, I have a page where I can check the status of all of the ones I'm involved in). When you want to toss it back to me, you'd do `:haircut:`.

Good point about existing things being marked NSFW from words in the title though, let me confirm they'd definitely get un-marked if we remove that logic. If that's the case, I think I'd rather run a backfill and automatically turn the `over_18` flag on all of them on retroactively, I'd prefer not to add another data attribute to `Link` for this. I'll look into that now.

13steinj:
Okay the emojis make a lot more sense now, thanks.

From what I can tell, yes they would, since currently the only thing marking them nsfw is that property; and their `over_18` is False. But it wouldn't hurt to double check.

Deimos:
So yeah, you're right that this would cause all things that are currently tagged NSFW due to their title to get untagged. I wasn't 100% sure since `nsfw_str` is inside `Link._defaults`, which should imply that it might actually get changed somewhere, but it turns out that's just never used at all (and should be removed from there).

So then this would be my suggestion about the best way to do this:

1. Roll out a commit does nothing except starts setting `over_18` on the link inside `_submit` if the title matches the regex. This will start setting `over_18` properly on any new links, but will also keep doing the check at display-time for now, so we don't untag all those old links.
2. Do a backfill on all the pre-existing links and set `over_18` to true if their title matches the regex, this will fix all the old links.
3. Once that's finished, roll out another commit that stops doing the check at display-time.

Does that make sense?

13steinj:
As long as backfill means over time / at once a lot / all of said submissions would get some kind of backup so the `over_18` attribute can be set off of that, yes makes sense and I can change this PR to do 1 instead soon.

13steinj:
@Deimos I made the changes you requested (:hair_cut: ?)

r2/r2/models/link.py
```diff
@@ -217,6 +217,12 @@ def _submit(cls, is_self, title, content, author, sr, ip,
             url = content
             selftext = cls._defaults["selftext"]
 
+        nsfwcheck = re.compile(r"\bnsf[wl]\b", re.I)
```
Deimos: No need to define another copy of the regex, you can just use `cls._nsfw` instead of this.

r2/r2/models/link.py
```diff
@@ -217,6 +217,12 @@ def _submit(cls, is_self, title, content, author, sr, ip,
...
+        over_18 = False
+
+        if bool(nsfwcheck.search(title)):
```
Deimos: Really minor thing, but since you're going to have to change this line anyway, you can drop the `bool()` conversion. Just `if cls._nsfw.search(title):` is good.

r2/r2/models/link.py
```diff
@@ -230,6 +236,7 @@ def _submit(cls, is_self, title, content, author, sr, ip,
             ip=ip,
             comment_tree_version=cls._choose_comment_tree_version(),
             is_self=is_self,
+            over_18=over_18
```
Deimos: Super tiny style nit: we make sure to put a trailing comma on the end of each line, so that if we need to add another new argument it just requires adding the new line, not also adding the comma on the previous line (like how you didn't need to do that here).

Deimos:
Really tiny stuff, thanks for switching to this way. So you can just create a new commit with the fixes and push that up separately without touching the existing commit. That makes it easier for people to review, since they can just look at what you changed, instead of needing to look at everything from scratch again. Then once things are approved you can squash them at that point (or I can just do it before I deploy).

By the way, your haircut emoji didn't actually trigger our stuff correctly, because you did the form of it with the underscore, just : haircut : is needed.

:nail_care:

13steinj:
@Deimos sorry! I apparently didn't get the notification or you sent it during the time I couldn't access github last week and kept getting unicorns. Saw this just now, and fixed. :haircut:

Deimos:
Looks correct, thanks. I'll test to make sure this is going to work correctly, then look at deploying this and doing a backfill to set all the old posts with matching titles to over18 as well. :fish:

13steinj:
:+1:

Deimos:
Deployed, thanks again: https://github.com/reddit/reddit/commit/156672bed621650a2bf2485f7ff4189b449ff891

PR #1420, opened 2015-10-04
Force successor permissions

13steinj:
When a top mod removes him/herself, there was no insurance to the fact that the next mod should recieve full permissions. This commit makes a common /r/redditrequest automated and should reduce strain on that subreddit. This also automates the removal of spam/deleted users if they would be the successor.

I've made the two properties in subreddit.py just because there are other semi-issues that rely on what would be provided by those properties, so for the sake of making it easier on future issues I was going to fix, I made them properties instead of determining them in api.py

### DIFF
--- a/r2/r2/controllers/api.py
+++ b/r2/r2/controllers/api.py
@@ -699,6 +699,9 @@ class ApiController(RedditController):
 
         """
         if container and container.is_contributor(c.user):
+            if container.is_moderator_with_perms(c.user, 'access'):
+                ModAction.create(container, c.user, 'removecontributor',
+                                 target=c.user, details='remove_self')
             container.remove_contributor(c.user)
 
 
--- a/r2/r2/controllers/api.py
+++ b/r2/r2/controllers/api.py
@@ -850,7 +850,9 @@ class ApiController(RedditController):
 
         # Log this action
         if new and type in self._sr_friend_types:
-            ModAction.create(container, c.user, action, target=victim)
+            details = (c.user == victim and 'remove_self') or None
+            ModAction.create(container, c.user, action, target=victim,
+                             details=details)
 
         if type == "friend" and c.user.gold:
             c.user.friend_rels_cache(_update=True)
--- a/r2/r2/models/subreddit.py
+++ b/r2/r2/models/subreddit.py
@@ -537,11 +537,43 @@ class Subreddit(Thing, Printable, BaseSite):
 
         return super(Subreddit, self).add_moderator(user, **kwargs)
 
+    def _handle_mod_succession(self):
+        from r2.models import ModAction
+        # cycle through each available moderator
+        for user_id, perms in self.moderators_with_perms().iteritems():
+            successor = Account._byID(user_id)
+            # skip deleted
+            if successor._deleted:
+                continue
+            # successor has full perms, end loop
+            if perms.is_superuser():
+                return
+            # if the successor does not have all perms, give it to him/her
+            successor_rel = self.get_moderator(successor)
+            successor_rel.set_permissions(perms.loads(None))
+            successor_rel._commit()
+            # Log this
+            ModAction.create(
+                self,
+                c.user,
+                'setpermissions',
+                target=successor,
+                details='set permissions on successor',
+                description=successor_rel.encoded_permissions
+            )
+            # end the loop as soon as a successor has been made
+            return
+
     def remove_moderator(self, user, **kwargs):
         hook = hooks.get_hook("subreddit.remove_moderator")
         hook.call(subreddit=self, user=user)
 
+        top_mod_id = self.moderators[0]
+
         ret = super(Subreddit, self).remove_moderator(user, **kwargs)
+
+        if user._id == top_mod_id and self.moderators:
+            self._handle_mod_succession()
 
         is_mod_somewhere = bool(Subreddit.reverse_moderator_ids(user))
         if not is_mod_somewhere:
--- a/r2/r2/models/subreddit.py
+++ b/r2/r2/models/subreddit.py
@@ -568,7 +568,7 @@ class Subreddit(Thing, Printable, BaseSite):
         hook = hooks.get_hook("subreddit.remove_moderator")
         hook.call(subreddit=self, user=user)
 
-        top_mod_id = self.moderators[0]
+        top_mod_id = self.moderators[0] if self.moderators else None
 
         ret = super(Subreddit, self).remove_moderator(user, **kwargs)
 


umbrae:
:eyeglasses: @Deimos

r2/r2/models/subreddit.py
```diff
@@ -553,6 +553,20 @@ def remove_moderator(self, user, **kwargs):
     def moderators(self):
         return self.moderator_ids()
 
+    @property
+    def top_moderator_account(self):
+        if len(self.moderators) >= 1:
+            return Account._byID(self.moderators[0])
+        else:
+            return FakeAccount()
```
13steinj: Just realized you guts may want this line to actually return `None` instead of a `FakeAccount()`

r2/r2/models/subreddit.py
```diff
@@ -553,6 +553,20 @@ def remove_moderator(self, user, **kwargs):
...
+
+    @property
+    def top_moderator(self):
+        if len(self.moderators) >= 1:
+            return self.get_moderator(self.top_moderator_account)
+        else:
+            return FakeSRMember(ModeratorPermissionSet)
```
13steinj: Same here

r2/r2/controllers/api.py
```diff
@@ -795,6 +800,86 @@ def check_api_friend_oauth_scope(self, type_):
...
+                potentialsuccessor = Account._byID(user_id)
+                potentialsuccessor_as_mod = sr.get_moderator_invite(potentialsuccessor)
+                potentialsuccessor_perms = potentialsuccessor_as_mod.get_permissions()
+                # uninvite each next successor if said successor is deleted / marked as spam
+                if potentialsuccessor._deleted or potentialsuccessor._spam:
+                    if potentialsuccessor._deleted:
+                        badsuccessordescription = "user is [deleted]"
+                    elif potentialsuccessor._spam:
+                        badsuccessordescription = "user is shadowbanned"
+                    sr.remove_moderator(potentialsuccessor)
```
13steinj: Whoops. Should be `sr.remove_moderator_invite`

r2/r2/controllers/api.py
```diff
@@ -795,6 +800,86 @@ def check_api_friend_oauth_scope(self, type_):
...
+                        sr,
+                        c.user,
+                        'uninvitemoderator',
+                        target=potentialsuccessor,
+                        details='account is not a valid successor',
+                        description=badsuccessordescription
+                    )
+                    continue
+                # if the successor does not have all perms, give it to him/her
+                if not potentialsuccessor_perms['all']:
```
13steinj: I didn't use `has_permissions` here nor on the mod cycle because I couldn't get it to work here. Weirdly enough this syntax worked fine.

r2/r2/controllers/api.py
```diff
@@ -795,6 +800,86 @@ def check_api_friend_oauth_scope(self, type_):
                 # this request.
                 self._auth_error(403, "insufficient_scope")
 
+    def _handle_mod_successorship(self, sr):
```
13steinj: You may want these cycles being just one, the list to be cycled through determined via `getattr`

r2/r2/controllers/api.py
```diff
@@ -870,6 +931,11 @@ def POST_unfriend(self, nuser, iuser, container, type):
         if type in ("friend", "enemy") and container != c.user:
             abort(403, 'forbidden')
         fn = getattr(container, 'remove_' + type)
+
+        # if the top mod is removing themself
```
13steinj: This comment is slightly inaccurate: should be "if the top mod is removing themself or an admin is removing them"

r2/r2/controllers/api.py
```diff
@@ -752,6 +754,9 @@ def POST_leavecontributor(self, container):
         """
         if container and container.is_contributor(c.user):
             container.remove_contributor(c.user)
+            if container.is_moderator_with_perms(c.user, 'access'):
```
Deimos: This seems unrelated, probably shouldn't be in here (or at least in a separate commit). I'm also not really sure why it's being added, why do we want a mod log entry when mods remove themselves as a contributor, when there isn't one for any other user removing themselves?

Deimos:
I think this change ended up being way over-complicated. There's no need to have moderators being auto-removed as part of this process, and I think the situation where the only moderator steps down while having a pending invite out with non-full permissions is probably so rare that it's really not worth worrying about.

Can we simplify it down to do *only* this: "if a top moderator leaves (or is removed), give full permissions to the first non-deleted moderator below them on the list"? It should also be done in the `remove_moderator()` function inside the `Subreddit` class, not in the individual API endpoints.

:nail_care:

r2/r2/controllers/api.py
13steinj: @Deimos The reasoning I had was that when mods remove anyone or themselves from being a mod, a mod action was created; and there's also a mod action made when removing themselves as a contributor if they remove themselves from the /about/contributor page (as that itself is done via `POST_unfriend` and not `POST_leavecontributor`)

13steinj:
@Deimos :haircut: 

I didn't remove the remove_self just in case, and left a comment [here](https://github.com/reddit/reddit/pull/1420#discussion_r44602352). If needed I can remove it (and or rebase from master to resolve merge conflicts)

r2/r2/controllers/api.py
Deimos: Alright. I don't think it's really a big deal, but either way it needs to be in a separate commit since it's completely unrelated to the "automatic permissions to successor" stuff.

r2/r2/controllers/api.py
```diff
@@ -870,6 +873,7 @@ def POST_unfriend(self, nuser, iuser, container, type):
         if type in ("friend", "enemy") and container != c.user:
             abort(403, 'forbidden')
         fn = getattr(container, 'remove_' + type)
+
```
Deimos: Nit: please remove the newline being added here. We try to avoid making whitespace changes except when we're already adjusting code for another reason, and this is separated from the actual changes.

r2/r2/controllers/api.py
```diff
@@ -883,7 +887,16 @@ def POST_unfriend(self, nuser, iuser, container, type):
                 contributor='removecontributor',
                 muted='unmuteuser',
             ).get(type, None)
-            ModAction.create(container, c.user, action, target=victim)
+            details = None
+            if (c.user == victim and
```
Deimos: This `if` is extremely difficult to follow (and `is` shouldn't be used to check string equality).

I really don't think it's even necessary, I'd probably just do `if c.user == victim and c.site.is_moderator(c.user):` and set the `remove_self` note on that. I can't see any way that the distinction of "moderator that has permissions to remove others is removing themselves" vs. "user is a moderator that doesn't have permissions to remove others, but is leaving this status" is important at all.

r2/r2/models/subreddit.py
```diff
@@ -553,11 +580,25 @@ def remove_moderator(self, user, **kwargs):
...
+        if len(self.moderators) > 0:
+            return self.get_moderator(self.top_moderator_account)
+
     def moderators_with_perms(self):
         return collections.OrderedDict(
             (r._thing2_id, r.get_permissions())
             for r in self.each_moderator())
 
+    @property
+    def moderator_invites(self):
```
Deimos: This property isn't used any more, and can be removed.

r2/r2/models/subreddit.py
```diff
@@ -553,11 +580,25 @@ def remove_moderator(self, user, **kwargs):
     def moderators(self):
         return self.moderator_ids()
 
+    @property
+    def top_moderator_account(self):
```
Deimos: Let's just drop both of these properties too, I think. We can just check the IDs directly above, so we don't really need `top_moderator_account` or `top_moderator`.

r2/r2/models/subreddit.py
```diff
@@ -540,6 +540,33 @@ def remove_moderator(self, user, **kwargs):
         hook = hooks.get_hook("subreddit.remove_moderator")
         hook.call(subreddit=self, user=user)
 
+        if user == self.top_moderator_account and len(self.moderators) > 1:
```
Deimos: This gets into some pretty heavy indenting, I think this is going to be a lot cleaner as a separate function. I'd just do something like this:

    # if the top mod is being removed, make sure the new one has full perms
    if user._id == self.moderators[0]:
        self._handle_mod_succession()

(There's no need to check that there's more than 1 moderator, that will be automatically handled by the list of potential successors being empty)

Then move the code inside here to a new function, which will save some indentation and make it a little easier overall.

r2/r2/models/subreddit.py
```diff
@@ -540,6 +540,33 @@ def remove_moderator(self, user, **kwargs):
...
+            from r2.models import ModAction
+            fullperms = self.moderator_permission_class.loads('+all')
+            # cycle through each available moderator
+            for user_id in self.moderators[1:]:
```
Deimos: It might be a bit easier to use `for user_id, perms in self.moderators_with_perms():` here, then you don't have to get the rel/perms manually each time.

Deimos: Note that to make this work most easily, you should move the call to this new successor function **below** the point in `remove_moderator()` where `ret = super(Subreddit, self).remove_moderator(user, **kwargs)` is called (so that the user being removed is no longer a mod, and won't be considered as a candidate, and you don't have to worry about skipping them or anything).

r2/r2/models/subreddit.py
```diff
@@ -540,6 +540,33 @@ def remove_moderator(self, user, **kwargs):
...
+                successor = Account._byID(user_id)
+                # skip deleted / shadowbanned users
+                if successor._deleted or successor._spam:
```
Deimos: I think we had probably better not skip `._spam` users for now. We still more or less consider them "valid mods" for most things (they're still in the mod list and such), so if we skip over them there's a risk of a weird case where a temporarily-shadowbanned user gets skipped as a successor, and full perms actually pass to the third (or lower) mod instead of the second. I think just skipping deleted ones is best to avoid that case for now.

r2/r2/models/subreddit.py
```diff
@@ -540,6 +540,33 @@ def remove_moderator(self, user, **kwargs):
...
+                    continue
+                successor_rel = self.get_moderator(successor)
+                successor_perms = successor_rel.get_permissions()
+                # if the successor does not have all perms, give it to him/her
+                if not successor_perms['all']:
```
Deimos: Just to make this a little simpler (and reduce indentation more), when you're doing this in the new function you can just do something more like:

    # if the successor already has full perms, we don't need to do anything
    if successor_perms["all"]:
        return

And then the code for setting permissions and creating the mod log entry doesn't need to be inside an `if:`.

Deimos:
Mostly just organization suggestions, looks pretty good overall. :nail_care:

13steinj:
@Deimos :haircut: should I rebase to resolve merge conflicts as well as squash the extra commits together?

r2/r2/models/subreddit.py
```diff
@@ -536,12 +536,46 @@ def add_moderator(self, user, **kwargs):
 
         return super(Subreddit, self).add_moderator(user, **kwargs)
 
+    def _handle_mod_succession(self):
+        if len(self.moderators) == 0:
+            return
```
13steinj: This check isn't actually needed, as the for loop would just do nothing instead...but from something I asked a month or so ago on stackoverflow, iterating over an empty list would still take up some processing power and an extremely small amount of time, so I added it anyway.

13steinj:
@Deimos :haircut: I've re-updated since I viewed this earlier today and noticed some things were inefficient.

bsimpson63:

PR #1422, opened 2015-10-05
Resolve issues with contest mode upvote only subs

13steinj:
Resolves the issues underlined in [this post](https://www.reddit.com/r/ModSupport/comments/3n9676/contest_mode_could_really_use_a_few_minor_ui/).

* This makes it so that the first issue no longer exists, as the score isn't being altered server side, but rather downvotes are blocked.

* This resolves the second issue in the same way.

* This resolves the third issue by making it so that a message similar to that which happens on an archived post's comments when voting on them (but of course, only for downvotes).

* This also resolves an issue noone knew about (as far as I can tell); that the comments could never be properly collapsed.

Brought to my attention by @1point618

Edit:

This adds:

* A message on the contest mode notification itself

### DIFF
--- a/r2/r2/controllers/api.py
+++ b/r2/r2/controllers/api.py
@@ -2199,6 +2199,11 @@ class ApiController(RedditController):
         if thing._age > thing.subreddit_slow.archive_age:
             store = False
 
+        if (dir < 0 and isinstance(thing, Comment):
+            if (thing.link_slow.contest_mode and
+                thing.subreddit_slow.contest_mode_upvotes_only):
+                store = False
+
         dir = (True if dir > 0
                else False if dir < 0
                else None)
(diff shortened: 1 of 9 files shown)


1point618:
I haven't actually seen this in action, but from discussions with @13steinj and cursory :eyes: on the code, it seems like it would fix all my issues and more.

13steinj:
@1point618 I'd take a video from my VM but my VM's screen resolution makes things weird.

1point618:
No worries, I trust ppl smarter than me to test it :smile:

13steinj:
Well if you want to test it yourself just get on ubuntu / a vm, go through the install guide, `git fetch` my branch run make in /r2, and test at your leisure.

13steinj:
@1point618 I'll need to update this in order to work for [this recent change](https://github.com/reddit/reddit/compare/bde2b1fffa71f28856f1978e2cee739044064686...d9daaaf39c0f3dc80ab6c80bf3938e23e3176e57), but it's [not done yet](https://www.reddit.com/r/redditdev/comments/3p8zxd/was_the_change_in_the_voting_javascript_closed/cw5a5p4), so what I have may break yet again.

13steinj:
I've refactored to work with the recent updates to the voting JS and rebased from the master branch

umbrae:
Thanks for this, we're not ready to invest more time in upvote only contest mode just now so I think we need to close this one. If we come back to it I'll dig this up.

PR #1425, opened 2015-10-07
Mod log RSS: fix <link> tag not being added

j0be:


Pokechu22:
To clarify, this fixes the bug where the modlog didn't have a permalink in the feed, right?  (I can't find the /r/bugs post right now but I know it exists)

Deimos:
Looks correct, thanks. I'm going to replace the commit message when I merge it though - commit messages should generally explain what the purpose/reasoning of the change is, not just be a description of the actual code change. It's similar to how this is considered a bad comment inside code:

    # add 1 to counter
    counter += 1

So a better commit message for something like this would probably be more like "Mod log RSS: fix <link> tag not being added"

:fish:

13steinj:
@Deimos I'm not sure but I think [this](http://www.reddit.com/r/bugs/comments/3nqpd3/rmodaboutlog_doesnt_include_permalinks_in_the_rss_feed/cvrkh4o) also needs to be changed.

Deimos:
@13steinj: ah yeah, you're correct. Fish removed, :nail_care: !

j0be:
Yep. Thanks for pointing it out 13steinj

j0be:
Ugh. Ignore that close/open. Accidentally clicked it.

j0be:
And I'll try to leave a better description in the future. Sorry about that. First commit to you all. :P

13steinj:
@j0be You may want to test and or rebase to your master and squash the extra commits.

`git rebase -i master`

13steinj:
@j0be I went ahead and actually tested via a local install in `reddit_shell`. You can view the process of my testing in [this pastebin](https://thepb.in/5616eda8c38331).

Disregarding some but not all of the attribute errors caused by me forgetting to do some shit, it confims the suspiscion that I secretly had, which is that `thing.subreddit` didn't need to be altered to `thing.target.subreddit`, and that the target does not have a subreddit, as it in and of itself has not had the wrapped properties added to it.

Long story short, keep `thing.target.parent_link`(maybe? I don't know if that attribute error was significant, and I don't have the time to check right now), but definitely exchange `thing.target.subreddit` back to just `thing.subreddit`. 

Or it may be better to just stop worrying and exchange the `make_permalink` methods with `make_permalink_slow()`

Deimos:
It's probably easiest to just use `make_permalink_slow()`. So you'd just do:

    if isinstance(thing.target, (Comment, Link)):
        permalink = thing.target.make_permalink_slow()

13steinj:
^ That

umbrae:
I'm gonna close this since it's a bit stale and still needs updates, feel free to comment and I will reopen if you'd like to fix it.

PR #1428, opened 2015-10-10
Fix rare issue upon editing self posts

13steinj:
Fixes a rare issue that occurs when editing self posts, as reported [here](https://www.reddit.com/r/bugs/comments/3iszjl/when_you_edit_a_post_that_is_in_your_recently/)

I would have fixed the other issues regarding the incorrect template when running on the userpage, but I'm not fluent enough in js to do so.

### DIFF
--- a/r2/r2/controllers/api.py
+++ b/r2/r2/controllers/api.py
@@ -1897,8 +1897,8 @@ class ApiController(RedditController):
             )))
 
         wrapper = default_thing_wrapper(expand_children = True)
-        jquery(".content").replace_things(item, True, True, wrap = wrapper)
-        jquery(".content .link .rank").hide()
+        jquery("body>div.content").replace_things(item, True, True, wrap = wrapper)
+        jquery("body>div.content .link .rank").hide()
 
     @allow_oauth2_access
     @validatedForm(
--- a/r2/r2/controllers/api.py
+++ b/r2/r2/controllers/api.py
@@ -3072,8 +3072,8 @@ class ApiController(RedditController):
 
         wrapper = default_thing_wrapper(expand_children = True)
         w = wrap_links(thing, wrapper)
-        jquery(".content").replace_things(w, True, True)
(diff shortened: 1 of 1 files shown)


umbrae:
Wow, good find.

:eyeglasses: @madbook

madbook:
good catch! one small request, would you mind applying the same fix to https://github.com/reddit/reddit-public/blob/master/r2/r2/controllers/api.py#L3242-L3243, since it looks like it would suffer from the same issue?

long term I think I can nuke most of the `replace_things` code altogether - it's more complicated than it needs to be and apparently produces fun bugs like this!  Until I can find time to do that though, scoping the response to the main `.content` div seams reasonable.

:nail_care:

13steinj:
@madbook do you mean the POST_hide endpoint? Just to be sure.

13steinj:
@madbook I think I figured out what you meant? It was hard to tell, as the url you gave was broken, but after testing I found the issue existed on distinguishing as well. :haircut:

madbook:
Ah, yeah, that's what I meant. I'm guessing what happened is that some other commit was pushed up to master after I posted that link, which moved the line highlighting.  Thanks for getting that too, I'll let you know when I get this merged in. :fish:

13steinj:
:P  No problem!

umbrae:
cb128961732396ff885d73076c74fa8fad7091f3 and 86ef81d62a6eabf501a9ec2b60d8e041e7fc04df

Thank you!

P.S. You've already got one open sourcerer trophy so I probably shouldn't add more each time. Maybe I could just change the text to "A bunch of stuff!" or something?

13steinj:
@umbrae Yeah; adding things would probably make my trophy case look ugly. If you want to change the text to that and the url link to [this](https://github.com/reddit/reddit/commits/master?author=13steinj) to clarify on what a "bunch of stuff" is, go for it :P

PR #1431, opened 2015-10-10
Fix inconsistency in post json

13steinj:
While post's selftext now shows differences between '[removed]' and '[deleted]', as well as in comments, this opened a new inconsistency with the json response of posts, which is when a post is removed the selftext key is `''` instead of `[removed]`.

This commit resolves this inconsistency.


umbrae:
Looks good, I will test. Thanks for the fix.

:eyeglasses: @umbrae

umbrae:
8b87ab1cf81557a4a73d2b4098ee38212d29ecfb thanks!

PR #1435, opened 2015-10-12
Fix /hidden tab when votes are public

13steinj:
The /hidden tab (although, unusable) is mistakenly added publicly when the user has public votes enabled.

This commit resolves the above.


MelissaCole:
:fish: Thanks!

MelissaCole:
699067a

PR #1440, opened 2015-10-19
Add a bunch of listings for the use of mods

13steinj:
I hope that I've actually fully completed this. It's complete as far as I can tell, but I need to make a few line notes.

This PR would add a /about/[locked|contests|nsfw] listing, make the current /about/unmoderated listing work for comments, and make /comments sortable. I've also included a bugfix for [this](https://www.reddit.com/r/bugs/comments/3p952j/filtered_subreddits_comments_page_is_not_respected/). You may or may not want me to add a feature flag to some of these, and you also, if you want to push b751b5f , you may want to wait until deciding on the pull request that I will make very soon after this one.

PR #1441, opened 2015-10-19
Add a "bottom" and "worst" comments sort for moderators

13steinj:
Adds both sorts for comments for mods with the "posts" perm only, as asked several times for on /r/ideasfortheadmins and /r/modsupport. If this is wanted, you may want b751b5f of #1440 to be changed slightly so that mods could sort the comments there as well. 

One thing that I'm having trouble deciding on is if comments should be forced to be uncollapsed on these sorts, so for not I didn't make un-collapsement forced but if you want, I added it in the third commit.

E: I added the third commit

### DIFF
--- a/r2/r2/public/static/css/reddit.less
+++ b/r2/r2/public/static/css/reddit.less
@@ -448,8 +448,12 @@ ul.flat-vert {text-align: left;}
 .drop-choices.tabdrop {margin-top: 2px;}
 .dropdown-title.tabdrop { display: none }
 
-.drop-choices .choice.hidden {
+.drop-choices .choice.hidden,
+.drop-choices .choice.modsort {
     display: none;
+}
+.modsorts-on .drop-choices .choice.modsort {
+    display: block;
 }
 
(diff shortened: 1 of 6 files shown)


r2/r2/public/static/css/reddit.less
```diff
@@ -448,9 +448,13 @@ ul.flat-vert {text-align: left;}
 .drop-choices.tabdrop {margin-top: 2px;}
 .dropdown-title.tabdrop { display: none }
 
-.drop-choices .choice.hidden {
+.drop-choices .choice.hidden,
+.drop-choices .choice.modsort {
     display: none;
 }
+.modsorts-on .drop-choices .choice.modsort {
```
13steinj: I didn't use a `body:not(.modsorts-on) ...` block just because some versions of IE don't actually support it, and they are used widely enough to potentially be significant.

umbrae:
Thanks for the work, sadly though I don't think we'd like to add additional sorts right now.

PR #1442, opened 2015-10-19
Fix Incorecct Name Scheme in TimeMenu

13steinj:
Fixes an incorrect naming scheme in the timemenu, as the User overview and user /comments section have an incorrect name in the time menu ("links from"), when it should be either "things from" or "comments from"

If this and #1440 is excepted, you may want me to fix this (spelling?) error there as well, as I didn't make the change there.

### DIFF
--- a/r2/r2/controllers/listingcontroller.py
+++ b/r2/r2/controllers/listingcontroller.py
@@ -33,8 +33,8 @@ from r2.lib.filters import _force_unicode
 from r2.lib.jsontemplates import get_usertrophies
 from r2.lib.pages import *
 from r2.lib.pages.things import wrap_links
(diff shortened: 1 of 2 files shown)


umbrae:
Code looks good, but maybe we can use "posts and comments from" instead of "things from"?

:eyeglasses: @umbrae

13steinj:
@umbrae :haircut:

umbrae:
:fish:

dwick:
typo in commit message `Incorecct`

umbrae:
To be honest I noticed that but was going to fix it when I released. ;) @13steinj if you care to fix that, feel free

umbrae:
429b85b5ca8b0f0b3b135e0b6040904c691afebf Thanks!

PR #1444, opened 2015-10-19
Clarify [removed] and [deleted].

minecast:
Make these terms more explicit in order to reduce user confusion.

### DIFF
--- a/r2/r2/lib/jsontemplates.py
+++ b/r2/r2/lib/jsontemplates.py
@@ -690,7 +690,7 @@ class LinkJsonTemplate(ThingJsonTemplate):
             if not thing.expunged:
                 return safemarkdown(thing.selftext)
             else:
-                return safemarkdown(_("[removed]"))
+                return safemarkdown(_("[removed by moderator]"))
         elif attr == "archived":
             return not thing.votable
         return ThingJsonTemplate.thing_attr(self, thing, attr)
--- a/r2/r2/lib/jsontemplates.py
+++ b/r2/r2/lib/jsontemplates.py
@@ -685,7 +685,7 @@ class LinkJsonTemplate(ThingJsonTemplate):
             if not thing.expunged:
                 return thing.selftext
             else:
-                return ''
+                return '[removed by moderator]'
         elif attr == 'selftext_html':
             if not thing.expunged:
                 return safemarkdown(thing.selftext)
--- a/r2/r2/models/link.py
+++ b/r2/r2/models/link.py
@@ -1478,9 +1478,9 @@ class Comment(Thing, Printable):
                     # If removed by an admin or moderator, distinguish that
                     # from being deleted by the user.
                     if item._spam:
-                        item.body = '[removed]'
+                        item.body = '[removed by moderator]'
                     else:
-                        item.body = '[deleted]'
+                        item.body = '[deleted by author]'
 
             if focal_comment == item._id36:
                 extra_css += " border"
--- a/r2/r2/models/link.py
+++ b/r2/r2/models/link.py
@@ -761,7 +761,10 @@ class Link(Thing, Printable):
             if item.deleted and not c.user_is_admin:
                 item.author = DeletedUser()
                 item.as_deleted = True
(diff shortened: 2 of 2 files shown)


13steinj:
There's literally no point to this. It's understandable enough, and I'll bet you my life savings you won't be able to find someone who actually is confused, (not literally, since there are some individuals out there, but you know what I mean). Even if one was confused. It's extremely easy to explain.

Also, this is not done for submissions, you only edited the comment text.

minecast:
To the average user, "removed" and "deleted" sound pretty much the same, I mean unless you explain it to them, most aren't going to know the difference. It is perfectly acceptable to say "The user removed the comment" or "The moderator deleted the comment". I don't really see any harm in doing it.

Also, I fixed that thing you mentioned.

13steinj:
It sounds nothing alike to me, nor to the "average user". Prove it. I see no harm, but I also see no gain.

Also...you didn't fix it for posts. Look at the logic. That only occurs if a post is both deleted and removed. And, even if you did it logically sound, doing it in `add_props` would break some technical things that are related to the use of the `expunged` attribute and it's boolean logic. To edit the text on posts, you'd need to make a change at least around [here](https://github.com/reddit/reddit/pull/1431/files), maybe somewhere else. But not in `add_props` AFAIK.

Pokechu22:
> It sounds nothing alike to me, nor to the "average user". Prove it. I see no harm, but I also see no gain.

I'll agree with @minecast -- it does seem rather similar.  Unless you see two next to each other, you'll have trouble noticing the difference.

It's not necessarily the average user that'll have an issue (or particularly care), it's more important to make it clear to those who would assume bad faith if they saw a missing comment.  Fine, some of those users will know the difference, but it's not obvious.

If a brand new user saw a "removed" comment, they wouldn't know what that means in comparison to "deleted".  All of us involved in this conversation understand the significance, but most people won't.

13steinj:
@Pokechu22 *shrug*. Never actually seen an issue come up regarding it or it's "confusion" in all my time.

r2/r2/lib/jsontemplates.py
```diff
@@ -685,7 +685,7 @@ def thing_attr(self, thing, attr):
             if not thing.expunged:
                 return thing.selftext
             else:
-                return ''
+                return '[removed by moderator]'
         elif attr == 'selftext_html':
             if not thing.expunged:
                 return safemarkdown(thing.selftext)
```
13steinj: @minecast You forgot to do it on #L693

english06:
Great pull request. Only mods really know the difference between "removed" and "deleted". This is a great clarification.

minecast:
@english06 Hmm, yeah. I'll try adding that when I can.

english06:
I am sadly the wrong guy to ask. Poking @13steinj for help.

13steinj:
@english06 Sorry, that's extremely strange. Both times you commented on something of mine / poked me I didn't get a notification.

@minecast No, it wouldn't, as `c.user_is_admin` is the state of the currently authenticated user being an admin, not the user that made the removal be an admin. You can not distinguish between admin and moderator removals, from a glance. Regardless, the admins never remove anything.

Pokechu22:
@13steinj Well, they may occasionally remove things, but that's usually due to a DCMA notice and thus is logged in [/r/ChillingEffects](//reddit.com/r/ChillingEffects) and is [already marked as such](https://www.reddit.com/r/ChillingEffects/comments/35urvq/test_post_please_ignore/).  I agree that it doesn't seem too necessary to distinguish between admin and moderator in that message, though.

13steinj:
@Pokechu22 yeah, I just disregarded the DMCA removals as those are completely different.

english06:
I certainly agree that they most likely rarely, if ever, act in that manner. I just wanted to account for it if possible. If not, I think the "removed by moderator" vs "deleted by user" clarification is very valuable in itself.

minecast:
Closing this and starting new PR with all changes in 1 PR due to my stupidity

PR #1445, opened 2015-10-20
Fixed unnecessary duplicate if

rubymeow:
literally zero priority - was just having a peek and this really bugged me -.-


Deimos:
This is not a correct fix, it tries to set the account ID above on the line:

    account_id = account_by_payingid(paying_id)

and only returns if that fails. You've changed it to return even if that lookup succeeds.

rubymeow:
Ahh wow that completely went over me... thanks for pointing the line out :)

PR #1446, opened 2015-10-20
Added CSS filter effects to allowed functions/attributes

rubymeow:
Filter effects are documented here: https://developer.mozilla.org/en-US/docs/Web/CSS/filter

Though they're still technically "experimental", it'd be nice to be able to work with them in subreddit CSS.

Excluded URL to prevent 3rd party request generation.


Pokechu22:
`url` actually properly is validated -- if you try to use a full URL it will be rejected; it needs to match the regex `"\A%%([a-zA-Z0-9\-]+)%%\Z`, and is used to specify an image: `background-image: url(%%logo%%);`.  

I do think allowing `filter` would make sense though.

13steinj:
I wanted to work with the filter property before and when I asked on /r/csshelp long ago someone told me it's use was specifically rejected. I *hope* that that was misinformation and it just wasn't thought of. Playing with said effects can make for nice styles.

rubymeow:
@Pokechu22 in this context it's for SVG filters so I'd assume it wouldn't work the same :-;

@13steinj interesting - I can't see any particular reason reddit would disallow new CSS3 elements (and to be honest there's a hell of a lot more missing) - the tone of the comment at the top of the file seems to put it that way

```
Parse and validate a safe subset of CSS.

The goal of this validation is not to ensure functionally correct stylesheets
but rather that the stylesheet is safe to show to downstream users.  This
includes:

    * not generating requests to third party hosts (information leak)
    * xss via strange syntax in buggy browsers

Beyond that, every effort is made to allow the full gamut of modern CSS.
````

13steinj:
Yeah, but there are specific things that are disallowed, such as importing fonts and such because of security reasons. Some other properties as well, there is always a reason.

rubymeow:
@13steinj yeah which is why I removed the URL reference (as it's already in there) - none of the other attributes/functions pose a security risk unless someone has a browser specific 0day for them... but even still that's just as likely to happen to any other function/attribute (and isn't very likely at all).

spladug:
Hey all, thanks for the patch but we can't accept `filter` at this time.  Here's my explanation from last time this came up:

https://github.com/reddit/reddit/pull/1058#issuecomment-76466180

unfortunately, IE9 usage is still high enough for us to not kill subreddit styles for those users.

GarethPW:
@spladug Is this still the case?

spladug:
ya

PR #1448, opened 2015-10-22
Add referrer_url  to the vote generic event

KeyserSosa:
This has the side effect of adding the referer as available `context` data for other generic events. 

cc @jophuds

### DIFF
--- a/r2/r2/lib/eventcollector.py
+++ b/r2/r2/lib/eventcollector.py
@@ -123,12 +123,6 @@ class EventQueue(object):
             for name, value in private_data.iteritems():
                 event.add(name, value)
 
-        if context_data:
-            referrer_url = context_data.get('http_referrer')
-            if referrer_url:
-                event.add("referrer_url", referrer_url)
-                event.add("referrer_domain", domain(referrer_url))
-
         self.save_event(event)
 
     @squelch_exceptions
@@ -365,11 +359,6 @@ class EventQueue(object):
             if thing_id36:
                 event.add("thing_id", int(thing_id36, 36))
 
-            referrer_url = request.headers.get('Referer', None)
-            if referrer_url:
-                event.add("referrer_url", referrer_url)
-                event.add("referrer_domain", domain(referrer_url))
-
         self.save_event(event)
 
     @squelch_exceptions
@@ -475,7 +464,12 @@ class EventV2(object):
 
         data["domain"] = request.host
         data["user_agent"] = request.user_agent
-        data["http_referrer"] = request.headers.get("Referer", None)
+
+        http_referrer = request.headers.get("Referer", None)
+        if http_referrer:
+            data["referrer_url"] = http_referrer
+            data["referrer_domain"] = domain(http_referrer)
+
         return data
 
     @classmethod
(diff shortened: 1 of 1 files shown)


KeyserSosa:
(I seem to be incapable of getting notifications right. Sorry if you got multiple highlights!)

:eyeglasses: @Deimos  @xiongchiamiov @spladug

jophuds:
Can we get the field name to be http_referrer (yes, I know it's not like that in http headers!).

r2/r2/lib/eventcollector.py
```diff
@@ -469,7 +475,7 @@ def get_context_data(self, request, context):
 
         data["domain"] = request.host
         data["user_agent"] = request.user_agent
-
+        data["referer"] = request.headers.get("Referer", None)
```
spladug: Joe got there first, but i was gonna ask about propagating the unfortunate spelling mistakes of our ancestors here :)

spladug:
:nail_care: looks great, just the field name

r2/r2/lib/eventcollector.py
KeyserSosa: I'm going to have to copy paste that.  I'm not sure my fingers are capable of adding an extra "r" at this point.  

(Will do!)

r2/r2/lib/eventcollector.py
```diff
@@ -123,6 +123,12 @@ def vote_event(self, vote, old_vote=None,
             for name, value in private_data.iteritems():
                 event.add(name, value)
 
+        if context_data:
+            referrer_url = context_data.get('referer')
+            if referrer_url:
+                event.add("referrer_url", referrer_url)
```
xiongchiamiov: If I'm thinking through this correctly, this means we'll end up with both `referrer_url` and `referer` (through being passed into `EventV2`'s constructor) being in these events, right?  Do we care about that duplication?

Alternatively, are we interested in having both `referrer_url` and `referrer_domain` available for every event (that is, by combining this block with the code down in `get_context_data()`?  I'm not sure what's desired from a data team perspective.

jophuds: This is a good point from @xiongchiamiov.

In quarantine events we get referrer_domain and the referrer_url as separate fields (although I'm not sure if these get populated). Seems like there are two options:

1. Send referrer_domain and referrer_url for just vote events
2. Send referrer_domain and referrer_url for all events

KeyserSosa: @jophuds @xiongchiamiov hrm.  I hadn't considered that as I hadn't noticed the constructor call.  I'll fix.

I'm inclined to just fix it to remove `http_referrer` and just have `referrer_url` and `referrer_domain` on all events when there's a `Referer` available.  This also means cleaning up the quarantine event to remove the resulting duplication.  I also don't feel strongly about it.

KeyserSosa:
I've added `referrer_url` and `referrer_domain` to `get_context_data` and therefore made it part of the defaults for all generic events.  

:haircut:

xiongchiamiov:
:fish: Cool, looks great to me.

spladug:
LGTM :fish:

PR #1449, opened 2015-10-25
Make the terms [deleted] and [removed] more explicit.

mistyhands:
For some users, these terms are seemingly interchangeable.


Pokechu22:
Just for reference, `git rebase -i master` would let you merge the changes together, and then they could be pushed with `git push --force` onto your local branch -- no need to start a new PR.  (And also, this was originally #1444 &mdash; there's some discussion there).

r2/r2/lib/jsontemplates.py
```diff
@@ -690,7 +690,7 @@ def thing_attr(self, thing, attr):
             if not thing.expunged:
                 return safemarkdown(thing.selftext)
             else:
-                return safemarkdown(_("[removed]"))
+                return safemarkdown(_("[removed by moderator]"))
```
13steinj: You forgot about L688

13steinj:
@Pokechu22 Given the fact that he made three commits instead of one, I assume he's working in browser, not via git; there's no rebase tool via the browser.

r2/r2/lib/jsontemplates.py
minecast: @13steinj I figured I'd just leave it blank as it was before.

minecast:
@13steinj The other 2 commits were me being stupid, unrelated to this particular commit. But yeah, I'm a git noob.

umbrae:
We want to do this, but we'd like to be a little bit more comprehensive about it, and particularly denote when something was removed by an admin. Do you want to try to tackle that also?

minecast:
@umbrae I'd be glad to! However, I'm quite unfamiliar with the reddit source. Would item.moderator return the user who removed an item?

13steinj:
@minecast Not really in this case.

I'll send a PR to your branch when I can. It's determined via `.verdict` and the string that it is. And...There's A small bug in removals that I'm currently fixing regarding the order in which the text is determined, so it may be better to wait until that's reviewed, since otherwise if I send a patch to your branch first that bug would only get worse.

minecast:
@13steinj Thank you!

minecast:
@umbrae added admin distinction & rebased

13steinj:
@minecast from first glance that should work but you have some indent errors.

minecast:
@13steinj thanks, and i did fix the indent errors and rebase again, here's what it looks like to me: http://i.imgur.com/rGxFD8r.png

Pokechu22:
@minecast You've got tabs in the code.  Python doesn't like tabs (especially mixed with spaces) - that's going to cause things to freak out.  You'll need to replace the tabs here with spaces.

minecast:
@Pokechu22 Thanks, fixed.

Pokechu22:
@minecast You've still got some in jsontemplates.py, although the ones in link.py are fixed.

minecast:
@Pokechu22 Good catch, thanks. Also fixed.

minecast:
Since verdict can have more states than 'mod-removed' or 'admin-removed', was it a bad idea to use else?

minecast:
@umbrae thoughts on current state of this pr?

umbrae:
@minecast I like it. We're going through open source pulls next week, hopefully I'll be able to dive deeper then. I think this looks good though at first blush.

umbrae:
:eyeglasses: @ajacksified @dforsyth This will change the response body for removed links in the API. Any qualms on that?

@minecast We'll probably need to take this PR internally and run with it, primarily because we'll need to "feature flag" this and test it with beta or employees to check if it breaks any API clients/RES/etc. If you want to be super awesome, you could check out how feature flags works: https://github.com/reddit/reddit/blob/master/r2/r2/config/feature/README.md 

Otherwise we can take this in house and tackle it. Thanks for the effort!

ajacksified:
@umbrae no problem here, we'll just need to update our optimistic "set-text-to-[deleted]" code to the new proper "deleted" text.

bsimpson63:

PR #1423, opened 2015-10-05
Fix Flairs

13steinj:
Fixes DOM updates as originally reported [here](https://www.reddit.com/r/bugs/comments/3l0yd2/colourchanging_link_flairs/).

PR #1450, opened 2015-10-25
Make mods able to customize flairs

13steinj:
As requested on /r/ModSupport:

Allows mods with the flair perm to customize the class of flairs. Also, if the user is a mod, always have a default template to choose from.

This also corrects a small UI bug which can be reproduced as mentioned [here](https://redd.it/3q3z17), as well as another tiny UI issue regarding spacing of the submit button, and the status span.

Edit: Before considering this, please consider #1423, as that fixes a bug, and if that gets merged, a minor change would have to be made here as well.

### DIFF
--- a/r2/r2/controllers/api.py
+++ b/r2/r2/controllers/api.py
@@ -3729,7 +3729,27 @@ class ApiController(RedditController):
             if not c.user_is_admin and not link.can_flair_slow(c.user):
                 abort(403)
 
+            subreddit = link.subreddit_slow
             link.set_flair(text, css_class, set_by=c.user)
+
+            # XXX: more gross UI code
+            # Push some client-side updates back to the browser.
+            jquery('.id-%s .entry .linkflairlabel' % link._fullname).remove()
+            title_path = '.id-%s .entry > .title > .title' % link._fullname
+
+            # TODO: move this to a template
+            classes = ' '.join('linkflair-' + c for c in css_class.split())
+            flair = format_html('<span class="linkflairlabel %s">%s</span>',
+                                classes, text)
+
+            if subreddit.link_flair_position == 'left':
+                jquery(title_path).before(flair)
+            elif subreddit.link_flair_position == 'right':
+                jquery(title_path).after(flair)
+
+            # TODO: close the selector popup more gracefully
+            jquery('body').click()
+
         else:
             if form.has_errors("name", errors.BAD_FLAIR_TARGET):
                 return
@@ -3743,6 +3763,8 @@ class ApiController(RedditController):
                 jquery.redirect('?name=%s' % user.name)
                 return
 
+            # XXX: more gross UI code
+            # Push some client-side updates back to the browser.
             wrapped_user = WrappedUser(
                 user, force_show_flair=True, include_flair_selector=True)
             rendered = wrapped_user.render(style='html')
--- a/r2/r2/lib/pages/pages.py
+++ b/r2/r2/lib/pages/pages.py
@@ -3997,6 +3997,7 @@ class FlairSelector(CachedTemplate):
         if link:
             flair_type = LINK_FLAIR
             target = link
+            target_type = "link"
             target_name = link._fullname
             attr_pattern = 'flair_%s'
             position = getattr(site, 'link_flair_position', 'right')
@@ -4009,6 +4010,7 @@ class FlairSelector(CachedTemplate):
         else:
             flair_type = USER_FLAIR
             target = user
+            target_type = "user"
             target_name = user.name
             position = getattr(site, 'flair_position', 'right')
             attr_pattern = 'flair_%s_%%s' % c.site._id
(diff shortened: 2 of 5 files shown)


r2/r2/public/static/css/reddit.less
```diff
@@ -1031,7 +1031,9 @@ a.banned-user { color: red; }
 }
 
 .flairselector .customizer { display: inline-block; }
-.flairselector .customizer input { display: none; }
+.flairselector .customizer input,
+.flairselector .mod.customizer span { display: none; }
+.flairselector .mod.customizer span { color: black; }
```
13steinj: Not really needed; but it makes things look nicer for the user flair selector.

r2/r2/lib/pages/pages.py
```diff
@@ -4029,6 +4031,11 @@ def __init__(self, user, site, link=None):
         else:
             choices = []
 
+        if admin:
+            admintempl = FlairTemplate()
+            admintempl.text = "ModFlair"
+            choices.append(target_wrapper(admintempl))
```
13steinj: This is essentially, a "FakeFlairTemplate"; and there is no worry about it's id being none; as this only occurs for mods / admins; and those inputs never require an ID.

r2/r2/controllers/api.py
```diff
@@ -3743,6 +3763,8 @@ def POST_flair(self, form, jquery, user, link, text, css_class):
                 jquery.redirect('?name=%s' % user.name)
                 return
 
+            # XXX: more gross UI code
+            # Push some client-side updates back to the browser.
```
13steinj: Left as a reminder like in POST_selectflair

r2/r2/templates/flairselector.html
```diff
@@ -31,26 +31,48 @@
               li_class += ' texteditable'
           if choice.flair_template_id == thing.matching_template:
               li_class += ' selected'
+          choice.flair_css = choice.flair_css_class if thing.target_type == "link" else choice.flair_css_class[6:]
```
13steinj: User flairs and their css classes are weird; and that's why this is needed.

r2/r2/templates/flairselector.html
```diff
@@ -31,26 +31,48 @@
...
+      <div class="flairselection">
+        <div class="flairremove">
+          (<a href="javascript://void(0)">${_('remove flair')}</a>)
+        </div>
+      </div>
+      <input type="hidden" name="flair_template_id">
+      <div class="customizer">
+        <input type="text" size="16" maxlength="64" name="text">
+      </div>
+      </br>
```
13steinj: Without this </br>; there is a UI issue regarding overflow when the status is displayed.

wodim: You probably meant `<br/>`.

13steinj: Yup. Thanks for the catch. I'll rebase at home.

umbrae:
Can you explain more the intent behind this and some examples of how it works? Screenshots also would be handy if you have them.

(Also just as an aside, we're pretty strict nowadays about not letting in code with TODO or XXX blocks in them. Probably don't fix them just yet so that we can determine if we think this is something we need, but if this is something that looks good we'll probably want to tackle those.)

Thanks for your contributions as usual!

13steinj:
@umbrae 

The intent behind this was a very recent /r/ModSupport post. As mods, there are times when flairs are needed to be mod only, so that no user can access them. While this is extremely easy for text alone, as a mod with flair perms can edit any flair's text; it is not easy for subreddit design and style.

To make mod only link flairs, as of right now, mods have to set up a comment tagging system using automoderater. It's good, but also is a bit of an inconvenicence. And manual selection of css class is it's own issue.

The base of the pr is this: change the endpoint to used to /post/flair, and add the proper code that would make client side updates there. This way, any mod with the flair perm can edit flairs' css class with no hassle whatsoever.

The added lines by the note [here](https://github.com/reddit/reddit/pull/1450#discussion_r42954010) make a flair template available to mods no matter what, even if there are no flairs, and allows for a blank slate css class wise and a semi blank slate text wise (text is generically "ModFlair").

Of course; this is just the basics. If you'd like, which you probably would since the blank slate flair is kinda done weirdly, I can expand this to actually make mod only flair templates in general, and make an entirely new selection underneath the non-mod's flair selection, which would definitely look nicer though I don't know how / if it should follow the limit on the amount of possible flair templates (which IIRC is 304? I can't find the exact line atm)

Taking screenshots is probably going to be difficult but should be doable. I'll get back when I clean a few small things up elsewhere.

umbrae:
Thanks for the contribution - at this stage I think this isn't a priority for us and we don't have enough time to think through its implications, so I'm going to close this.

I know it's a bummer to write a patch this large and have it closed on you, sorry about that.

13steinj:
@umbrae Nah it's cool :P

PR #1451, opened 2015-10-25
Move generation of S attribute to builder.py

13steinj:
Moves the generation of the S attribute to builder.py, previously in link.py's Comments' `add_props`.

This change makes the S attribute consistent with all the others, and will make it show up in all the places that it is supposed to, such as the inbox, which was the reason why [this bug report was originally filed on /r/bugs](https://redd.it/3oi9n6)

The order of the attributes are kept as they were.

### DIFF
--- a/r2/r2/models/builder.py
+++ b/r2/r2/models/builder.py
@@ -109,6 +109,10 @@ class Builder(object):
                 if sr.can_ban(user):
                     can_ban_set.add(sr_id)
 
+        # hasattr check so no errors occur for non comments
+        links = Link._byID(set(l.link_id for l in items if hasattr(l, 'link_id')),
+                           data=True, return_dict=True, stale=self.stale)
+
         #get likes/dislikes
         try:
             likes = queries.get_likes(user, items)
@@ -173,6 +177,12 @@ class Builder(object):
             except AttributeError:
                 pass
 
+            if hasattr(item, "sr_id") and item.sr_id is not None:
+                w.subreddit = subreddits[item.sr_id]
+
+            if hasattr(item, "link_id") and item.link_id is not None:
+                w.link = links[item.link_id]
+
             if isinstance(item, Comment):
                 if not hasattr(w, 'subreddit'):
                     w.subreddit = item.subreddit_slow
@@ -207,9 +217,6 @@ class Builder(object):
                            {"user": w.author.name}),
                     link="/user/%s" % w.author.name,
                 )
-
-            if hasattr(item, "sr_id") and item.sr_id is not None:
-                w.subreddit = subreddits[item.sr_id]
 
             w.likes = likes.get((user, item))
 
--- a/r2/r2/models/builder.py
+++ b/r2/r2/models/builder.py
@@ -173,6 +173,15 @@ class Builder(object):
             except AttributeError:
                 pass
 
+            if isinstance(item, Comment):
+                if not hasattr(w, 'subreddit'):
+                    w.subreddit = item.subreddit_slow
+                if not hasattr(w, 'link'):
+                    w.link = item.link_slow
+                if item.author_id == w.link.author_id and not w.link._deleted:
+                    add_attr(w.attribs, 'S',
+                             link=w.link.make_permalink(w.subreddit))
+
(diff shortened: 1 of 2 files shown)


r2/r2/models/builder.py
```diff
@@ -173,6 +173,15 @@ def wrap_items(self, items):
             except AttributeError:
                 pass
 
+            if isinstance(item, Comment):
+                if not hasattr(w, 'subreddit'):
+                    w.subreddit = item.subreddit_slow
+                if not hasattr(w, 'link'):
+                    w.link = item.link_slow
+                if item.author_id == w.link.author_id and not w.link._deleted:
```
13steinj: I got weird results with just assuming the attributes were there; sometimes they were, sometimes they weren't; so just in case they are not, they'd be built.

umbrae:
This looks reasonable, however we should probably be pulling those items in batch before the loop so that it's a bit quicker. You can see [here](https://github.com/reddit/reddit/pull/1451/files#diff-af03c6f59a55db9d7946ad8013d50ffcR104) that we're actually already pulling in subreddits in this way.

Does that make sense?

13steinj:
@umbrae I'm sorry, but I'm having a difficulty understanding. Do you mean move the addition of the `.subreddit` and `.link` to L139-ish?

umbrae:
Oh sorry, it also looks like my link didn't work.

Essentially using the `_slow` properties can be, well, slow. So we generally try to avoid doing them in a loop.

You can see here that we're actually loading the subreddits in batch before the loop: https://github.com/13steinj/reddit/blob/S-in-inbox/r2/r2/models/builder.py#L104

We'd probably want to do something similar with links.

Does that make more sense?

r2/r2/models/builder.py
```diff
@@ -109,6 +109,10 @@ def wrap_items(self, items):
                 if sr.can_ban(user):
                     can_ban_set.add(sr_id)
 
+        # hasattr check so no errors occur for non comments
+        links = Link._byID(set(l.link_id for l in items if hasattr(l, 'link_id')),
+                           data=True, return_dict=True, stale=self.stale)
```
13steinj: I used `stale=self.stale`, but in the current method via the `add_props`, `stale=True` no matter what. So you may want to double check this, just in case. Under casual observation I've had no issue.

r2/r2/models/builder.py
```diff
@@ -199,9 +218,6 @@ def wrap_items(self, items):
                     link="/user/%s" % w.author.name,
                 )
 
-            if hasattr(item, "sr_id") and item.sr_id is not None:
-                w.subreddit = subreddits[item.sr_id]
-
```
13steinj: Had to be moved up to L180 because underneath that is where the attribute is inserted.

r2/r2/models/builder.py
```diff
@@ -173,6 +177,21 @@ def wrap_items(self, items):
...
+                w.subreddit = subreddits[item.sr_id]
+
+            if hasattr(item, "link_id") and item.link_id is not None:
+                w.link = links[item.link_id]
+
+            if isinstance(item, Comment):
+                if not hasattr(w, 'subreddit'):
+                    w.subreddit = item.subreddit_slow
+                if not hasattr(w, 'link'):
+                    w.link = item.link_slow
```
13steinj: My reasoning for leaving in these `hasattr` checks are this: While it *should* work normally via lines L180-L184, in that same respect, these checks are done currently via the `add_props` method. Are they actually needed? I don't know. I haven't run into any problems with them not being in here, but just in case I left them.

13steinj:
@umbrae Done :P. You may want to read the line notes I made just in case, because the nuances invoked my own curiosity, so I left said notes in case I made an error. :haircut:

13steinj:
@umbrae I don't like bumping, but it just looks as if this was missed, so :haircut: and should I rebase to resolve merge conflicts?

umbrae:
No worries about rebasing, and you're right, I did miss this. I'll review.

13steinj:
:P 

@umbrae but what about 91f492f ?

bsimpson63:

PR #1454, opened 2015-11-01
Threaded modmail parameter

13steinj:
This allows one to get modmail in it's threaded form via the API, which can be useful depending on how the user works.

(Of course, without having to constantly enable the user preference if they for whatever reason want flat modmail)

I've also added two different commits for a GUI to select threaded modmail via the browser. If you'd like the first, just disregard the second. If you'd like the second, squash the third commit with the second. If you'd like neither, disregard both :P.

E: of course, the line in example.ini would have to be moved to reddit's run.ini

### DIFF
--- a/r2/example.ini
+++ b/r2/example.ini
@@ -873,3 +873,4 @@ feature_upgrade_cookies = off
 feature_multireddit_customizations = off
 # Test if `https_testing_img_test` is loaded with an acceptable HTTPS cert according to users' browsers
 feature_test_https_certs = off
+feature_threaded_modmail = {"url": "threaded_view"}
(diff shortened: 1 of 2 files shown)


umbrae:
Thanks for the contribution - I think we need to close this so that we can have a larger discussion about threaded modmail in general, though.

Also just as an aside this isn't the way we intend for feature flags to be used: they're typically for short-to-medium term things that need to go out to a segment of people. In this case we'd need to keep this feature flag around forever, which would have deleterious effects on code quality and also having a bunch of flags like that ends up being not great for zookeeper performance.

PR #1457, opened 2015-11-06
Add Images to /about/stylesheet for non-mods

13steinj:
Adds images to /about/stylesheets for easy view, even when a non mod, since the images were already accessible via inspect element as well as via the JSON template, as semi-wanted on /r/csshelp numerous times. Copypasta is beautiful :P

### DIFF
--- a/r2/r2/lib/pages/pages.py
+++ b/r2/r2/lib/pages/pages.py
@@ -2812,20 +2812,26 @@ class CreateSubreddit(Templated):
             )
 
 
-class SubredditStylesheet(Templated):
-    """form for editing or creating subreddit stylesheets"""
-    def __init__(self, site = None,
-                 stylesheet_contents = ''):
-        allow_image_upload = site and not site.quarantine
+class SubredditStylesheetBase(Templated):
+    """Base subreddit stylesheet page."""
+    def __init__(self, stylesheet_contents, **kwargs):
         raw_images = ImagesByWikiPage.get_images(c.site, "config/stylesheet")
         images = {name: make_url_protocol_relative(url)
                   for name, url in raw_images.iteritems()}
-
-        Templated.__init__(
-            self,
-            site=site,
+        super(SubredditStylesheetBase, self).__init__(
             images=images,
             stylesheet_contents=stylesheet_contents,
+            **kwargs
+        )
+
+
+class SubredditStylesheet(SubredditStylesheetBase):
+    """form for editing or creating subreddit stylesheets"""
+    def __init__(self, site=None, stylesheet_contents=''):
+        allow_image_upload = site and not site.quarantine
+        super(SubredditStylesheet, self).__init__(
+            stylesheet_contents=stylesheet_contents,
+            site=site,
             allow_image_upload=allow_image_upload,
         )
 
@@ -2872,14 +2878,10 @@ class SubredditStylesheet(Templated):
                 w.gilded_message = "this comment was fake-gilded"
         return wrapped.render(style="html")
 
-class SubredditStylesheetSource(Templated):
+
+class SubredditStylesheetSource(SubredditStylesheetBase):
     """A view of the unminified source of a subreddit's stylesheet."""
-    def __init__(self, stylesheet_contents):
-        raw_images = ImagesByWikiPage.get_images(c.site, "config/stylesheet")
-        images = {name: make_url_protocol_relative(url)
-                  for name, url in raw_images.iteritems()}
-        Templated.__init__(self, images=images,
-            stylesheet_contents=stylesheet_contents)
+
 
 class AutoModeratorConfig(Templated):
     """A view of a subreddit's AutoModerator configuration."""
--- a/r2/r2/lib/pages/pages.py
+++ b/r2/r2/lib/pages/pages.py
@@ -2875,7 +2875,11 @@ class SubredditStylesheet(Templated):
 class SubredditStylesheetSource(Templated):
     """A view of the unminified source of a subreddit's stylesheet."""
     def __init__(self, stylesheet_contents):
-        Templated.__init__(self, stylesheet_contents=stylesheet_contents)
+        raw_images = ImagesByWikiPage.get_images(c.site, "config/stylesheet")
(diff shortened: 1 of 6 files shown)


r2/r2/public/static/css/highlight.css
```diff
@@ -8,7 +8,7 @@
 }
 
 /* github.com style (c) Vasily Polovnyov <vast@whiteants.net> - from the highlight.js source */
-pre {
+pre.subreddit-stylesheet-source {
```
13steinj: Nuance due to how this style would affect the pre elements showing the image urls.

r2/r2/templates/subredditstylesheetsource.html
```diff
@@ -33,3 +33,31 @@
...
+              ${name}
+            </b>
+            <br/>
+            <span>link:</span>
+            <pre class="img-url">url(%%${name}%%)</pre>
+            <br/>
+          </div>
+        </li>
+        </%def>
+      ${make_li(prototype=True)}
```
13steinj: I don't understand why this is here, as it adds a empty image both here and in subredditstylesheet.html, which seems unnecessary. Just in case I left it in both,

umbrae:
@13steinj Do you think you could provide a screenshot if you're able? I know you said getting screenshots was kinda tricky for you before.

:eyeglasses: @madbook do you think you could take a peek at this at some point?

13steinj:
@umbrae I was actually able to fix the bug I had with screenshots around a week ago, so [here](http://i.imgur.com/N3HDZRm.png), just don't ask about the subreddits in the top bar :P

r2/r2/lib/pages/pages.py
```diff
@@ -2875,7 +2875,11 @@ def rendered_comment(comments, gilded=False):
 class SubredditStylesheetSource(Templated):
     """A view of the unminified source of a subreddit's stylesheet."""
     def __init__(self, stylesheet_contents):
-        Templated.__init__(self, stylesheet_contents=stylesheet_contents)
+        raw_images = ImagesByWikiPage.get_images(c.site, "config/stylesheet")
+        images = {name: make_url_protocol_relative(url)
+                  for name, url in raw_images.iteritems()}
+        Templated.__init__(self, images=images,
+            stylesheet_contents=stylesheet_contents)
```
madbook: I think we ought not to duplicate this.  It might be better to move this into a base class that `SubredditStylesheet` and `SubredditStylesheetSource` both inherit from, e.g. something like

```python
class SubredditStylesheetBase(Templated):
    def __init__(self, stylesheet_contents, **kwargs):
        raw_images = ImagesByWikiPage.get_images(c.site, "config/stylesheet")
        images = {name: make_url_protocol_relative(url)
                  for name, url in raw_images.iteritems()}
        super(SubredditStylesheetBase, self).__init__(
            images=images,
            stylesheet_contents=stylesheet_contents,
            **kwargs
        )

class SubredditStylesheet(SubredditStylesheetBase):
    def __init__(self, site=None, stylesheet_contents=""):
        allow_image_upload = site and not site.quarantine
        super(SubredditStylesheet, self).__init__(
            stylesheet_contents=stylesheet_contents,
            site=site,
            allow_image_upload=allow_image_upload,
        )

class SubredditStylesheetSource(subredditStylesheetBase):
    pass
```

r2/r2/public/static/css/highlight.css
madbook: I think we should avoid editing this file since it's 3rd party stuff.  It really should be moved to `css/lib` or something like that.  I know it s a little annoying, but can you do the inverse (i.e add styles to _undo_ this file's effect) in `reddit.less`?

r2/r2/templates/subredditstylesheetsource.html
madbook: the empty `li` is used as a template on the upload form, [here](https://github.com/reddit/reddit/blob/master/r2/r2/templates/subredditstylesheet.html#L137-L151). It doesn't really need to be on this page, but as with the above, I'd rather not duplicate this.  If `SubredditStylesheet` and `SubredditStylesheetSource` are changed to inherit from a base class, then this could be moved into a function in that base template rather than being repeated in both files.

madbook:
A couple of suggestions to get rid of the code duplication, but otherwise good! Thanks! :nail_care:

13steinj:
@madbook I'm haircutting just in case the @mention won't pick up :haircut: 

Sure thing; will make all the changes regarding duplication ASAP.

However; is it even possible to "undo" the style in `reddit.less`? IIRC `highlight.js` is called and (injected if that is the right term) which puts the `highlight.css` rules at a higher priority than `reddit.less`'s and as such any rule in `reddit.less` can't overwrite a `highlight.css` rule. Unless I'm being silly in my reasoning.

madbook:
no problem, going to :nail_care: to move this back into the correct state.

Yes! CSS rule precedence is determined by a number of things, one of which _is_ stylesheet order.  So by default `highlight.css` will take precedence.  _However_, selector specificity has a more significant impact on specificity, so as long as the selector in `reddit.less` is more complicated than the one in `highlight.css`, the `reddit.less` rule should win out.

13steinj:
@madbook :haircut: & should I squash the first two commits?

The third commit makes it so that the empty blue stylesheet box does not display and instead the "there's nothing here" does if there is no stylesheet, in case wanted, since I just noticed it.

madbook:
hey @13steinj, sorry for letting this drag on forever.  I checked this out locally and found a couple of really small nits.  I've already fixed them up, so if you don't mind I'll just squash those into your commits (they'll still show up as yours when I merge them in).  [here's a patch file of the changes](https://gist.github.com/madbook/37052f88ec459b092b2f).  Sound ok?

13steinj:
@madbrook Yeah that's fine :P

madbook:
:fish: Cool, I'll get this merged in monday probably

madbook:
e2864b8104255fdf259643a4d292bc74cabb983b...05843ebce7910d73a730ebf7ad32d76079e1dc5c  Thanks!

PR #1461, opened 2015-11-08
Return an error on a duplication of a permban

13steinj:
Returns an error if the user being permwikibanned / permbanned is already permwikibanned / permbanned respectively. Originally reported [here](https://www.reddit.com/r/bugs/comments/3p2h8l/banned_twice_and_user_got_sent_two_messages/)


13steinj:
I'm closing because I just realized this is technically incomplete.

PR #1463, opened 2015-11-10
Don't mark the mute message for translation

13steinj:
When marked for translation the message would get sent in the mom's language. That shouldn't happen. Reported [here](http://www.reddit.com/r/bugs/comments/3s6uts).


MelissaCole:
:fish: Thanks! We had to recently do that for the [ban messages](https://github.com/reddit/reddit/commit/5a5db0899284412cd26a44501969fed792e55296) as well.

MelissaCole:
1be353d

PR #1464, opened 2015-11-12
Add '/u/reddit` to share-via-email placeholders

amici-ursi:
When you click the "share" link below a post, you have several options. One of them is to share it via email (http://imgur.com/8rLzrHP). You can use this to share posts with other redditors, but it's not obvious (https://www.reddit.com/r/ModSupport/comments/3sbpjr/why_cant_we_share_posts_with_other_redditor/). 

I suggest adding `/u/reddit` as one of the placeholders so sharers know usernames are an option.


umbrae:
We intentionally removed that from the language this time around because we were worried that new folks wouldn't understand what `/u/` means and would make errors. We're open to testing it down the line, but for now I think we're going to close this. Thanks for the contribution though!

PR #1466, opened 2015-11-14
GUI improvements for the force add moderators section for admins

13steinj:
Admins can forcibly add moderators to subreddits with their own tool. However; the error fields that this commit removes were duplicated (because they are already retrieved from the `<%call>` to [`userlisting.html`](https://github.com/reddit/reddit/blob/master/r2/r2/templates/userlisting.html#L79)), and when an admin encounters said error(s), it shows up twice because of this.

This stops the error from appearing twice.

This also adds a permissions form for admins to use if they wish to do so, as well as makes the ALREADY_MODERATOR error actually return when admins forcibly add mods. 

While the last commit probably wouldn't have real use in reddit itself, there are potential "look and feel" improvements (excess mod log entries), and can help those that wish to run their own reddit instance.

### DIFF
--- a/r2/r2/controllers/api.py
+++ b/r2/r2/controllers/api.py
@@ -1036,7 +1036,8 @@ class ApiController(RedditController):
         else:
             permissions = None
 
-        if type == "moderator_invite" and container.is_moderator(friend):
+        if (type in ("moderator_invite", "moderator") and
+                container.is_moderator(friend)):
             c.errors.add(errors.ALREADY_MODERATOR, field="name")
             form.set_error(errors.ALREADY_MODERATOR, "name")
             return
--- a/r2/r2/controllers/api.py
+++ b/r2/r2/controllers/api.py
@@ -1163,6 +1163,11 @@ class ApiController(RedditController):
             table = jquery("." + type + "-table").show().find("table")
             table.insert_table_rows(user_row, index=index)
             table.find(".notfound").hide()
+            # hide the listing in the invite table if it exists
+            if type == 'moderator':
+                table = jquery(".moderator_invite-table")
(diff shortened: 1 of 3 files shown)


tdohz:
:eyeglasses: @Deimos

r2/r2/controllers/api.py
```diff
@@ -1062,7 +1062,8 @@ def POST_friend(self, form, jquery, friend,
         else:
             permissions = None
 
-        if type == "moderator_invite" and container.is_moderator(friend):
+        if (type in ("moderator_invite", "moderator") and
+           container.is_moderator(friend)):
```
Deimos: Style nit: wrong indentation, should be 8 spaces in from the `if` line so it looks like:

```
if (type in ("moderator_invite", "moderator") and
        container.is_moderator(friend)):
    c.errors.add(errors.ALREADY_MODERATOR, field="name")
```

r2/r2/controllers/api.py
```diff
@@ -1175,6 +1175,10 @@ def POST_friend(self, form, jquery, friend,
             table = jquery("." + type + "-table").show().find("table")
             table.insert_table_rows(user_row, index=index)
             table.find(".notfound").hide()
+            if type == 'moderator':
```
Deimos: Can you add a comment above this saying what it's doing? Doing UI code in here is ugly, so it's nice to see what it's for from comments instead of needing to puzzle it out.

Deimos:
Just a couple of really minor things, code seems good. :nail_care:

13steinj:
@Deimos :haircut: sorry, didn't see this in my messages, was just going through the listing. Fixed.

bsimpson63:

PR #1472, opened 2015-11-21
Don't translate the gold gift message.

13steinj:
Translating this message both makes no sense for the recipient, and can potentially be harmful to an anonymous buyer's privacy (think small subreddit and you know a user there speaks a certain language that is rare for that subreddit's user base, or something similar, point is it's potentially harmful in unique cases). Hopefully, the last commit retranslates the message (I don't know how the `_()` method works fully) that is sent to the buyer, as translating that would make sense.


yowar:
https://twitter.com/morrisganis

r2/r2/controllers/ipn.py
```diff
@@ -590,9 +590,9 @@ def finish(self, parameters, txn_id,
                         "supports the site and makes future development "
                         "possible. For example, one month of reddit gold "
                         "pays for 5 instance hours of reddit's servers.")
-            message += "\n\n" + strings.gold_benefits_msg
+            message += "\n\n" + _(strings.gold_benefits_msg)
```
MelissaCole: These are [already translated](https://github.com/reddit/reddit/blob/master/r2/r2/lib/strings.py#L124). Variables can't be translated. The way this works (I think) is the string inside `_()` lets [crowdin](https://www.reddit.com/r/i18n) know that it can be translated, so it enters the string into its database so that it can be translated by users.

MelissaCole: Oh I just saw that you untranslated it down below. Hmm might need to see if there's a way to untranslate it or just create a new variable.

13steinj: Whoops. Out of curiosity can `_('hi %s bye' % data)` be properly translated?

If it can't be untranslated there's only two strings that would need to have both a translated and untranslated version (lounge_msg and gold_benefits_msg)

MelissaCole: Yup, but we like to specify the variable name (like [here](https://github.com/reddit/reddit/pull/1472/files#diff-1c750c2158fff6e35fa4c6cb1aee45d7R123)) so that the placeholder is clearly spelled out in crowdin.

In crowdin, `_('%(number)s days until thanksgiving' % dict(number=2))` will let the translator know that there is a placeholder for `number` so that they can move that to the appropriate spot in the translation.

13steinj: I'm extremely confused as to why I got a github notification that seems to no longer exist on Github itself about some kind of art twitter (morris + ganis?), but thanks :P

Pokechu22: @13steinj You might want to look at [this comment](https://www.reddit.com/r/OutOfTheLoop/comments/3lj1ez/whats_the_story_behind_create_morrisganis/cv6z0pv) by /u/13steinj.

13steinj: @Pokechu22 HAA HAA. I know *what* it is. I don't know why someone named 'morris ganis' made the note nor how it decided to stop existing within the 3 seconds it took me to press "respond on Github"

kemitche: re: i18n stuff: You would want to translate formattable strings like so:

`_("%(number)s days until Thanksgiving") % dict(number=2)`. The formatting should be *outside* of the `_()` call for everything to work properly.

MelissaCole: Ah yeah that's right. Thanks, Keith! :)

MelissaCole:
:fish: (I'm going to exclude 211398420fbfdaac476cb3d5ee59ae7a42ba8057). Thanks!

MelissaCole:
https://github.com/reddit/reddit/commit/528cfcf867604d9f7389126b729740506075f04a

PR #1474, opened 2015-11-21
Add subreddit creator to subreddit json template

13steinj:
Anybody can view the subreddit's creator via the creator text on the sidebar. Somewhat surprisingly, this is not included in the JSON response. This commit puts it in the JSON response.

### DIFF
--- a/r2/r2/controllers/front.py
+++ b/r2/r2/controllers/front.py
@@ -879,7 +879,8 @@ class FrontController(RedditController):
         Data includes the subscriber count, description, and header image."""
         if not is_api() or isinstance(c.site, FakeSubreddit):
             return self.abort404()
-        item = Wrapped(c.site, accounts_active_count=c.site.accounts_active)
+        item = Wrapped(c.site, author=c.site.author_slow,
+                       accounts_active_count=c.site.accounts_active)
         Subreddit.add_props(c.user, [item])
         return Reddit(content=item).render()
 
--- a/r2/r2/lib/jsontemplates.py
+++ b/r2/r2/lib/jsontemplates.py
@@ -283,6 +283,7 @@ class SubredditJsonTemplate(ThingJsonTemplate):
         collapse_deleted_comments="collapse_deleted_comments",
         comment_score_hide_mins="comment_score_hide_mins",
         # community_rules="community_rules",
+        creator="author",
         description="description",
         description_html="description_html",
         display_name="name",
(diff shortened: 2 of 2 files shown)


r2/r2/lib/jsontemplates.py
```diff
@@ -397,6 +398,12 @@ def thing_attr(self, thing, attr):
                 return c.user.use_subreddit_style(thing)
             else:
                 return True
+        elif attr == 'author':
+            if not thing.author:
```
13steinj: Normally, this `elif` block wouldn't need to be added, and the subreddit creator's name would be retrieved via the `else` block. However, there are times that there is *no* author (or at least that seems to be the case from the SubredditInfoBar `creator_text` [double checking if the subreddit has an author](https://github.com/reddit/reddit/blob/c93f7b8c84895bf5c5df93516e1b47cb315484e4/r2/r2/lib/pages/pages.py#L1177) in the firstplace, and /r/askreddit not having a creator, so it isn't showing up because of [this similar check](https://github.com/reddit/reddit/blob/85f61036814a3f58ccdb4a4d85d83ce07c292425/r2/r2/templates/subredditinfobar.html#L109), just as an example), and unless I'm mistaken, the `ThingJsonTemplate`'s `thing_attr` does not have a check for items that have no author, and when the `name` attribute will be called on the `None` that the author would be, you'd get an AttributeError.

r2/r2/controllers/front.py
```diff
@@ -879,7 +879,8 @@ def GET_about(self):
         Data includes the subscriber count, description, and header image."""
         if not is_api() or isinstance(c.site, FakeSubreddit):
             return self.abort404()
-        item = Wrapped(c.site, accounts_active_count=c.site.accounts_active)
+        item = Wrapped(c.site, author=c.site.author_slow,
+                       accounts_active_count=c.site.accounts_active)
```
13steinj: This is the only place that the attribute is missing and needs to be added. In all other places (/subreddits listings), the builder takes care of it.

r2/r2/lib/jsontemplates.py
```diff
@@ -283,6 +283,7 @@ class SubredditJsonTemplate(ThingJsonTemplate):
         collapse_deleted_comments="collapse_deleted_comments",
         comment_score_hide_mins="comment_score_hide_mins",
         # community_rules="community_rules",
+        creator="author",
```
13steinj: The word 'creator' was used for the same reason the terminology on account suspensions was changed in the json response; publicly, it's the subreddit's creator, not the subreddit's author.

umbrae:
Thanks for this one, I think we're going to pass on it for now as I don't think it's a hugely necessary change.

PR #1476, opened 2015-11-22
Fix mod log json

13steinj:
Fixes an issue with the mod log json and rss feeds.

### DIFF
--- a/r2/r2/templates/modaction.xml
+++ b/r2/r2/templates/modaction.xml
@@ -29,7 +29,7 @@
 <%
   title = '%(subreddit)s: %(mod)s %(action)s'
   title %= {
-    'subreddit': thing.subreddit.name,
+    'subreddit': thing.sr_name,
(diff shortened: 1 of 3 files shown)


r2/r2/templates/modaction.xml
```diff
@@ -44,9 +44,9 @@
   if thing.details_text:
     title += ' (%s)' % thing.details_text
 
-  if isinstance(thing, Comment):
+  if  hasattr(thing, "target") and isinstance(thing.target, Comment):
```
13steinj: The below line, I *did* change `make_permalink_slow` to `make_permalink`, but it is not showing up. I assume you guys already did it in a commit that is not public, but, that was not all that needed to be done.

umbrae:
I think this is out of date now that we've moved to atom. Feel free to update (and explain what the issue was?) if it's still busted in atom.

PR #1479, opened 2015-11-26
inject_test_data: better subreddit generation

13steinj:
The amount of subreddits generated / found to have their links put into is now based on user specification and the current popular subreddits of reddit.com

Sorted by date to create the subreddits in the general order that the actual subreddits were created.

### DIFF
--- a/scripts/inject_test_data.py
+++ b/scripts/inject_test_data.py
@@ -143,6 +143,21 @@ def fetch_listing(path, limit=1000, batch_size=100):
         # see: https://github.com/reddit/reddit/wiki/API#rules
         time.sleep(2)
 
+def fetch_subreddits(modeler, num_subreddits):
+    if num_subreddits > 1000:
+        raise ValueError("Can't fetch more than 1000 subs")
+    srs = fetch_listing('/subreddits', num_subreddits)
+    sorted_srs = sorted(srs, key=lambda sr: sr['created_utc'])
+    models = []
+    extra_info = {}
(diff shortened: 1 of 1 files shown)


spladug:
The choice of three subreddits in there was to specifically exercise different kinds of posts which is useful in testing:

* pics: links to images, lots of thumbnails but not many embeds
* videos: links to videos, lots of embeds
* askhistorians: pure self text

it wasn't meant to be an actual simulation of reddit.com.

What's the goal of adding more subreddits to the mix?

13steinj:
@spladug the goal of adding more subreddits to the mix was just for the ability to have a wider variety of test data across the given subreddits. While it most definitely would take longer if people are say, passing the argument to be 100, I left the default at the small 10 (and I can definitely understand going smaller, but going to small may end up in needing a fallback to the previous method if the current popular subreddits aren't of the types you mentioned).

It's also not meant to simulate reddit.com, that's just how the names for the subs are generated because I couldn't think of a decent alternative that ended up making sense.

spladug:
Thanks for the patch, but I'm going to close this. I'd really like to avoid making the script take longer; especially with throwaway VMs like the new vagrant setup bootstrapping content should take as little time as we can make. I might even tune down how many links/comments get modeled to that end.  Most importantly in the time budget is that we get a useful cross-section of content types for testing with, and the best way to guarantee that is to fix the list of subreddits being used to a known good list.

PR #1482, opened 2015-11-28
Add /api/quarantine_opt[in|out] to apidocs

13steinj:
Adds /api/quarantine_optin and /quarantine_optout to the api docs, since it's kinda weird that they aren't there. In restrospect, you probably will want to add to the API rules that this should be done only under user supervision, like voting.

On a separate note, you make want to make it so that these functions will return an error if the subreddit isn't quarantined to begin with.


ChristoCoder:
good idea

bsimpson63:

Issue #1340, opened 2015-06-20
Wrong case


SoniEx2:
While most of reddit's text starts with a lowercase letter, there are 2 checkboxes on the sidebar that don't. They're the "Show this subreddit's theme" and "Show my flair on this subreddit" checkboxes.

13steinj:
This isn't an isse. Some things have appropiate capitalization, others don't. Can you give some major comparisons?

SoniEx2:
It appears that only this thing and the legal stuff have appropriate capitalization. The legal stuff I can understand but I believe this should be lowercased.

13steinj:
For bugs that are "believed"> go to [/r/bugs](http://reddit.com/r/bugs) and report *there*. *Here* is only for issues that are concrete, e.g., someone went in and recognized a code issue.

SoniEx2:
This is an actual issue. Excluding the legal stuff, there's only 1 place that has "proper capitalization". That place should be lowercased for consistency with the rest of the site. For example all preferences (on the prefs page) are lowercased, these are per-subreddit preferences, why shouldn't they be lowercased?

13steinj:
Consistencies of cases aren't issues. Let me give you a description from the /r/bugs sidebar.


>Bug tracker @ GitHub 
>
>PLEASE only use GitHub for confirmed bugs with the code -- NOT for questions or feature requests.

What you altered, isn't really code. If you think the capitalization issue is a bug, it should go on /r/bugs instead. You say that everything is lowercased for consistency? What about the "Sumbit a new link/post"? Some notices, and, more importantly, everything that is UPPERCASE by default?

If anything, this is an annoyance. It doesn't harm anything, anyone, etc. It belongs in /r/bugs.

SoniEx2:
/r/bugs is for hypothetical bugs. technically UPPERCASE is the same as lowercase but it's used only for headings.

Submit a new link/post is a per-subreddit configuration, and it's not a PREFERENCE NAME, unlike the flair/css toggles which ARE preference names.

There are actually only VERY FEW (about 5 or so) places which have proper case/title case and aren't legal stuff.

This is an **inconsistency**, not an annoyance.

13steinj:
Even if it were a bug;

>PLEASE only use GitHub for confirmed bugs with the code -- NOT for questions or feature requests.

This isn't an issue with code really.

[...]

PR #1483, opened 2015-12-02
Fixing casing inconsistency

sashaperigo:
I was looking for an easy first PR to submit, and thought this issue would be an easy one to tackle: https://github.com/reddit/reddit/issues/1340.

Seems like there was some debate over whether or not this was actually an issue that needed to be fixed, but I figured I'd open a PR just in case!

### DIFF
--- a/r2/r2/templates/flairprefs.html
+++ b/r2/r2/templates/flairprefs.html
@@ -28,7 +28,7 @@
         %endif
         >
     <label for="flair_enabled">
-        ${_("Show my flair on this subreddit. It looks like:")}
+        ${_("show my flair on this subreddit. it looks like:")}
     </label>
   </form>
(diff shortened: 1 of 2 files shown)


kjoconnor:
Thanks for this, but unfortunately we don't accept PRs like these as per our [contributing guidelines](https://github.com/reddit/reddit/blob/master/CONTRIBUTING.md):

> 6. Please don't submit minor tweaks to text on the site, such as changes to punctuation, spacing or wording. Any string changes will cause all previous translations of that string to other languages to be invalidated, so we try to only change strings when the difference is significant.

sashaperigo:
Ah, I read that bullet but wasn't sure if casing fell into that category or not. Thanks for the quick response!

Issue #1319, opened 2015-05-18
Missing operand - install script for Ubuntu


yakutsa:
I have tried to install github reddit using install script for Ubuntu. I am using Ubuntu 12.04 on VPS but I get the following error:

+ '[' '!' -d /home//src ']'
+ mkdir -p /home//src
+ chown /home//src
chown: missing operand after `/home//src'
Try `chown --help' for more information.

Why am I getting this error?

spladug:
It looks like `$REDDIT_USER` is not defined for some reason. It should come automatically from the `sudo` environment if you're running the script that way -- otherwise, you'll need to set it before running the script.

e.g.

```
$ sudo ./install-reddit.sh
```

or

```
# export REDDIT_USER=myusername
# ./install-reddit.sh
```

yakutsa:
@spladug , 

Yes, you are right.

Thank you for your great support.

PR #1484, opened 2015-12-09
FIX: #1319 discover user even when running without sudo

maximveksler:
I've just launched a fresh instance on Digital Ocean, these guys give you a Ubuntu instance with root ssh access via a private key. I've then ran the install script and sure enough it failed because of #1319 so yeah, had I've exported REDDIT_USER it would have worked, but why not fix it without me having to configure stuff? 

The suggested patch simply takes the value of USER if no SUDO_USER environment variable present. 

The installer is happier, and so is it's users.

### DIFF
--- a/install-reddit.sh
+++ b/install-reddit.sh
@@ -48,7 +48,8 @@ set -e
 # Configuration
 ###############################################################################
 # which user to install the code for; defaults to the user invoking this script
-REDDIT_USER=${REDDIT_USER:-$SUDO_USER}
+CURRENT_USER=${SUDO_USER:-$USER}
+REDDIT_USER=${REDDIT_USER:-$CURRENT_USER}
 
 # the group to run reddit code as; must exist already
 REDDIT_GROUP=${REDDIT_GROUP:-nogroup}


spladug:
If I'm understanding the digital ocean setup correctly, does that mean for you `$USER` would be `root` and reddit would be installed under the root account?

maximveksler:
@spladug yes. exactly. Having given that more though it would also happen had I've logged in as maximveksler and attempted to install reddit under normal account. 

 1. User under sudo - works.
 1. User defined REDDIT_USER - works.
 1. User logged in (root) - fails.
 1. User granted himself root permission and logged in - fails.

Case 3 is what happened to me, on the DigitalOcean case, my patch fixes that.

spladug:
I don't think it's a good idea for the reddit code (or any network service for that matter) to run as root, and I'd rather the install script not encourage that. Perhaps a better error message when the user is not set would work?

spladug:
I've merged e6de184151f36dce9402f0240be05425e1ecc1ec which should address the lack of user check.

PR #1490, opened 2015-12-30
Disable voting arrows on archived comments

caseif:
This PR is fairly straightforward, disabling the voting arrows on comments which have been archived. However, it does not impact the appearance of voting arrows on top-level posts, as they may take on a strange appearance when rendered as part of a listing of multiple posts given the disparity in formatting from non-archived posts.

### DIFF
--- a/r2/r2/templates/comment_skeleton.html
+++ b/r2/r2/templates/comment_skeleton.html
@@ -23,8 +23,8 @@
 <%inherit file="printable.html"/>
 <%namespace file="utils.html" import="plain_link" />
 
-<%def name="midcol(display=True, cls = '')">
-  ${parent.midcol(display=display, cls = cls)}
+<%def name="midcol(cls = '')">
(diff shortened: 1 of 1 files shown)


13steinj:
@caseif May I ask why exactly?

Just "disabling" it looks a little weird (at least for me). I personally find the current system of "when clicked, an error message pops up" a bit nicer.

(And no, I don't speak for reddit; was just wondering)

caseif:
@13steinj It should look fairly normal (I tested this by setting the `display` attribute to `none` on the actual site, which to my knowledge should produce the same effect. That aside, I believe hiding buttons provides a better UX than allowing them to be clicked only to display an error message explaining they're unusable.

13steinj:
Yeah, I mean I tested in the same way and to me personally it just looked a
little off. Call me old fashioned but I like the error message better. Just
my opinion :P

umbrae:
Thanks for the work, I think we prefer to have the votes visible even on archived comments though.

PR #1495, opened 2016-01-05
Add PROMOTE_COST_BASIS for reddit-plugin-adzerk

zeantsoi:
:eyeglasses: @spladug @dwick 

Per discussion in Slack, this resolves the `AttributeError` in _reddit-plugin-adzerk_.

I looked into adding this as a config var (parsed as [`choices`](https://github.com/reddit/reddit/blob/master/r2/r2/lib/configparse.py#L69)), but it'd lack the [`name`](https://github.com/reddit/reddit/blob/master/r2/r2/lib/utils/utils.py#L111) property that is invoked in code.

### DIFF
--- a/r2/r2/models/promo.py
+++ b/r2/r2/models/promo.py
@@ -39,6 +39,9 @@ from r2.models.subreddit import Subreddit, Frontpage
 PROMOTE_STATUS = Enum("unpaid", "unseen", "accepted", "rejected",
                       "pending", "promoted", "finished")
 
+PROMOTE_COST_BASIS = Enum('fixed_cpm', 'cpm', 'cpc',)
+
+
 class PriorityLevel(object):
     name = ''
     _text = N_('')


r2/r2/models/promo.py
```diff
@@ -39,7 +39,9 @@
 PROMOTE_STATUS = Enum("unpaid", "unseen", "accepted", "rejected",
                       "pending", "promoted", "finished")
 
-class PriorityLevel(object):
+PROMOTE_COST_BASIS = Enum('fixed_cpm', 'cpm', 'cpc',)
+
+
     name = ''
```
spladug: what's going on here? it looks like all these bits that used to be in the PriorityLevel class have been orphaned?

zeantsoi: That's me being bad at vi. Fixed.

dwick:
:fish:

spladug:
:fish:

PR #1475, opened 2015-11-21
Refactor POST_friend and show mods and mod invites their permissions / permission changes.

13steinj:
This PR fixes duplicate permbans/permwikibans, refactors POST_friend to be more efficient or return proper errors when needed, sends messages to those that become wikicontributors/wikibanned, sends a message to mods when they are demodded / their permissions altered, and shows moderator invites their permissions under the acceptance question so they know what they are getting into. 

Edit: if 350338c of this PR is accepted, 991abc3 of #1466 is not needed.

References:

* https://www.reddit.com/r/bugs/comments/3o9p3c/when_banning_adding_someone_to_the_wiki_as_a/?ref=search_posts

* https://www.reddit.com/r/bugs/comments/3o9oms/a_user_can_be_banned_from_a_subreddits_wiki_and/?ref=search_posts

* https://www.reddit.com/r/bugs/comments/3p2h8l/banned_twice_and_user_got_sent_two_messages/?ref=search_posts

* https://www.reddit.com/r/ModSupport/comments/3seyzg/if_a_moderators_permissions_are_changed_have_a_pm/?ref=search_posts

PR #1496, opened 2016-01-06
Non mod modmail

13steinj:
@bsimpson63 asked me to separate this from #1475 

I don't really know what logisitics for non mod modmail were wanted them so like in 1475 I just have separate commits with increasing changes. First commit fixes the bug; the rest modify behavior as necessary.

It can be argued that L2206 and L2316 can be slow, however; the python2 (not sure if 3, but 2 is what matters) interpreter does not evaluate functions in boolean operator based comparisons (probably using the wrong terminology >.>) until the statement before the operator and the operator itself are evaluated, eg:

    >>> setup = """
    ... from time import sleep
    ... def foo():
    ...     sleep(10)
    ...     return True
    ... """
    >>> from timeit import timeit
    >>> timeit("bool(False and foo())", setup=setup)
    1.9073486328125e-06
    >>> timeit("bool(True and foo())", setup=setup, number=1)
    10.010693073272705

So they are only used if the other two checks pass. Or all commits but the first may not even be necessary. I just did them to make the behavior "fit".

### DIFF
--- a/r2/r2/models/link.py
+++ b/r2/r2/models/link.py
@@ -1999,7 +1999,7 @@ class Message(Thing, Printable):
         # also, only global admins can be message spammed.
         if not skip_inbox and to and (not m._spam or to.name in g.admins):
             # if "to" is not a sr moderator they need to be notified
-            if not sr_id or not sr.is_moderator(to):
+            if not sr_id or not sr.is_moderator_with_perms(to, 'mail'):
                 # Record the inbox relation, but don't give the user
                 # an orangered, if they PM themselves.
                 # Don't notify on PMs from blocked users, either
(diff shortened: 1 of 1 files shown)


Pokechu22:
> It can be argued that L2206 and L2316 can be slow, however; the python2 (not sure if 3, but 2 is what matters) interpreter does not evaluate functions in boolean operator based comparisons (probably using the wrong terminology >.>) until the statement before the operator and the operator itself are evaluated, eg:

That's called [short-circuit evaluation](https://en.wikipedia.org/wiki/Short-circuit_evaluation) and is pretty standard.

13steinj:
What Pokechu22 said ^

bsimpson63:
I'm not comfortable merging this without understanding what the issue is. Do you have time/desire to look into it more?

13steinj:
@bsimpson63 Yes; I'll look into this further.

bsimpson63:
:running:

umbrae:
Closing for now as this looks to have been dropped.

PR #1502, opened 2016-01-10
Allow mods to edit automod's items if they have the posts permission

13steinj:
Allows the mods of a subreddit to edit automod's posts/comments if they have the "posts" permission, as asked for before several times, most recently [here](https://www.reddit.com/r/ModSupport/comments/402cgw/make_uautomoderator_posts_editable_by_any/).


umbrae:
Thanks for your time, but I don't think we want this. I think it breaks a fundamental expectation of how the site works by letting a user edit another user's content (even if a bot).

PR #1504, opened 2016-01-12
Saved category in json

13steinj:
Adds saved categories to the HTML and JSON response for reddit gold users. For better API and stuff. I don't know what I'd call it nor why I'm continuing typing at this point, just rambling on on the keyboard...

### DIFF
--- a/r2/r2/lib/jsontemplates.py
+++ b/r2/r2/lib/jsontemplates.py
@@ -750,6 +750,9 @@ class LinkJsonTemplate(ThingJsonTemplate):
     def raw_data(self, thing):
         d = ThingJsonTemplate.raw_data(self, thing)
 
+        if c.user.gold and thing.saved:
+            d["saved_category"] = thing.saved_category if thing.saved_category else None
+
         if c.permalink_page:
             d["upvote_ratio"] = thing.upvote_ratio
 
@@ -846,6 +849,9 @@ class CommentJsonTemplate(ThingTemplate):
             data["removal_reason"] = "legal"
         else:
             data["removal_reason"] = None
+
+        if c.user.gold and item.saved:
+            data["saved_category"] = item.saved_category if item.saved_category else None
 
         if not item.author._deleted:
             author = item.author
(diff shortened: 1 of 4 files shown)


r2/r2/lib/jsontemplates.py
```diff
@@ -750,6 +750,9 @@ def generate_image_links(preview_object, censor_nsfw=False):
     def raw_data(self, thing):
         d = ThingJsonTemplate.raw_data(self, thing)
 
+        if c.user.gold and thing.saved:
+            d["saved_category"] = thing.saved_category if thing.saved_category else None
```
13steinj: This if expression seemed nicer, since "no category" would be made None, instead of empty quotes.

r2/r2/templates/printable.html
```diff
@@ -101,6 +101,9 @@
       rowclass += " deleted"
     if hasattr(what, "saved") and what.saved:
       rowclass += " saved"
+    if c.user.gold and hasattr(what, "saved") and what.saved and hasattr(what, "saved_category"):
+      item_saved_category = "no-category" if not what.saved_category else what.saved_category
```
13steinj: It's impossible for category names to have a "-" in them, so this isn't really an issue. And it seemed nicer than " saved-" with no category suffix.

r2/r2/templates/printable.html
```diff
@@ -101,6 +101,9 @@
       rowclass += " deleted"
     if hasattr(what, "saved") and what.saved:
       rowclass += " saved"
+      if c.user.gold and hasattr(what, "saved_category"):
+        item_saved_category = "no-category" if not what.saved_category else what.saved_category
```
13steinj: I'm recommenting since I rebased for that green checkmark and to remove dupliace logical checks.

"no-category" seems better than just "saved-None" or "saved-null" or "saved-" or no rowclass added at all; and since categories can't have a - in them, there is no problem with confusing this with an actual category.

r2/r2/public/static/js/saved.js
```diff
@@ -160,11 +161,17 @@ r.saved.SaveButton = {
         var $category = $el.parents('.thing').find('.save-category').hide()
         $el.text(r._('save'))
         $el.thing().removeClass('saved')
+        $el.thing().attr('class', function(i, c) {
+            return (c.replace(/(^|\s)saved\S+/g, ''));
```
13steinj: Same trick as 58fd686

umbrae:
Thanks for this, I think we're going to wait on this and see if there's another way we can handle saved in the API down the line.

PR #1505, opened 2016-01-12
Add travis integration

KeyserSosa:
The only way to test travis is to deploy to master with a `.travis.yml` file and wait for travis-ci to pick up the change.  To do some pre-testing here, the `Vagrantfile` has been split into two distinct machines:

 * `default` is, as before, the full reddit build (we have an opportunity to rename this)
 * `travis` is a minimal install with only the packages and services required to run `nosetests` against the codebase.  It tries to mimic what will happen when we try to build this on travis-ci's workers.  

As part of this addition, I've moved `install-reddit.sh` to `install/reddit.sh` and populated that `install` folder with common scripts used for both `default` and `travis` builds.


KeyserSosa:
:eyeglasses: @bsimpson63 @spladug @ketralnis 

Most of the new shell scripts are just copy-paste from the original `install-reddit.sh`.  For the remaining bits, I used all ten of my thumbs.  

The sticky points were either around permissions or removing stock packages that travis installs by default.  Specifically:

 * There is a script to downgrade cassandra, since their default is 2.0
 * There are some `dpkg` and `apt` acropatics (aptobatics?) to uninstall their libmemcached and substitute our own.  

Most importantly: 

screenshot 2016-01-12 16 21 53

jedberg:
Continuous integration for reddit?  Did I just see flying bacon? :P

Vagrantfile
```diff
@@ -104,49 +104,66 @@ Vagrant.configure(2) do |config|
...
-  # additional setup
-  config.vm.provision "shell", inline: <<-SCRIPT
-    if [ ! -f /var/local/additional_setup ]; then
-      apt-get install -y ipython avahi-daemon
-      touch /var/local/additional_setup
-    else
-      echo "additional setup already run"
-    fi
-  SCRIPT
+  config.vm.define "default" do |redditlocal|
```
bsimpson63: note to self: will need to do `vagrant up default` to name the VM and ensure that this block runs

Vagrantfile
```diff
@@ -104,49 +104,66 @@ Vagrant.configure(2) do |config|
...
-      echo "install script already run"
-    fi
-  SCRIPT
-
-  # setup private code
-  if File.exist?("#{code_share_host_path}/private/vagrant_setup.sh")
-    config.vm.provision "shell",
-      path: "#{code_share_host_path}/private/vagrant_setup.sh",
-      args: [vagrant_user]
+  config.vm.define "travis" do |travis|
```
bsimpson63: Can you add a comment here or at the top of the file explaining that this is never run, but is to try to replicate what will happen in travis?

install/setup_postgres.sh
```diff
@@ -0,0 +1,53 @@
...
+IS_DATABASE_CREATED=$(sudo -u postgres psql -t -c "$SQL")
+
+if [ $IS_DATABASE_CREATED -ne 1 ]; then
+    cat <<PGSCRIPT | sudo -u postgres psql
+CREATE DATABASE reddit WITH ENCODING = 'utf8' TEMPLATE template0 LC_COLLATE='en_US.utf8' LC_CTYPE='en_US.utf8';
+CREATE USER reddit WITH PASSWORD 'password';
+PGSCRIPT
+fi
+
+sudo -u postgres psql reddit <  $FUNCTIONSQL
```
bsimpson63: Don't feel strongly, but you could move all the stuff in functions.sql into this file to avoid needing to pass the path. It doesn't look like anything else is relying on functions.sql.

bsimpson63:
Maybe not worth worrying about yet, but how will plugins work with this?

.travis.yml
```diff
@@ -0,0 +1,33 @@
...
+language: python
+
+python:
+    - "2.7"
+
+virtualenv:
+    system_site_packages: true
+
+services:
+    - cassandra
```
bsimpson63: Is it just much faster to include cassandra as a service rather than installing it in travis.sh, even with the downgrade?

install/setup_postgres.sh
KeyserSosa: This is really obvious in retrospect.  Fixing now!

install/travis_downgrade_cassandra.sh
```diff
@@ -0,0 +1,37 @@
...
+#
+# All portions of the code written by reddit are Copyright (c) 2006-2015 reddit
+# Inc. All Rights Reserved.
+###############################################################################
+set -x -e
+
+CASSPACKAGE=apache-cassandra-$CASSANDRA_VERSION
+
+sudo rm -rf /var/lib/cassandra
+wget http://archive.apache.org/dist/cassandra/$CASSANDRA_VERSION/$CASSPACKAGE-bin.tar.gz
```
bsimpson63: no way to use the datastax version?

.travis.yml
KeyserSosa: Good point.  I'll give it a try without this line to see if it can still attach to a by-hand allocated cassandra service.  I can't see why it wouldn't.

bsimpson63:
:nail_care: looks great, just a couple questions

KeyserSosa:
:haircut:

spladug:
To keep [cool URLs in docs etc.](https://www.w3.org/Provider/Style/URI.html) could `install-reddit.sh` still exist and just link to `install/reddit.sh`?

Vagrantfile
```diff
@@ -25,7 +25,15 @@
...
-# To start your vagrant box simply enter `vagrant up` from {ROOTDIR}/reddit.
+# This VagrantFile allows for the creation of two VMs:
+#   * default: the primary VM, with all services necessary to run reddit
+#              locally against the local codebase.
+#   * travis:  Testing-only VM suitable for running `nosetests` and debugging
+#              issues encountered without having to wait for travis-ci to pick
+#              up the build.  This will *not* be the same environment as
+#              travis, but it should be useful for repairing broken tests.
+#
+# To start your vagrant box simply enter `vagrant up default` from {ROOTDIR}/reddit.
```
spladug: Does this mean people will have to add "default" every time they want to up a default install? Could that be... defaulted for ergonomics?

install/done.sh
```diff
@@ -0,0 +1,88 @@
...
+# The Original Developer is the Initial Developer.  The Initial Developer of
+# the Original Code is reddit Inc.
+#
+# All portions of the code written by reddit are Copyright (c) 2006-2015 reddit
+# Inc. All Rights Reserved.
+###############################################################################
+set -e
+
+###############################################################################
+# Configuration
```
spladug: this doesn't really make sense as a configuration section anymore. these variable should be coming from elsewhere and the flow of just reading the script to find and modify them is gone.

install/install_services.sh
```diff
@@ -0,0 +1,65 @@
...
+postgresql-client
+rabbitmq-server
+haproxy
+nginx
+gunicorn
+sutro
+PACKAGES
+
+# we don't want to upgrade to C* 2.0 yet, so we'll put it on hold
+apt-mark hold cassandra || true
```
spladug: this should probably be in the cassandra script?

install/install_cassandra.sh
```diff
@@ -0,0 +1,45 @@
...
+
+APTITUDE_OPTIONS=${APTITUDE_OPTIONS:-"-y"}
+export DEBIAN_FRONTEND=noninteractive
+
+# add the datastax cassandra repos
+echo deb http://debian.datastax.com/community stable main | sudo tee /etc/apt/sources.list.d/cassandra.sources.list
+wget -qO- -L https://debian.datastax.com/debian/repo_key | sudo apt-key add -
+
+# grab the new ppas' package listings
+sudo apt-get update
```
spladug: it's kinda a bummer that we'll be calling `apt-get update` so many times now. that's quite slow.

install/install_apt.sh
```diff
@@ -0,0 +1,104 @@
...
+# The Original Developer is the Initial Developer.  The Initial Developer of
+# the Original Code is reddit Inc.
+#
+# All portions of the code written by reddit are Copyright (c) 2006-2015 reddit
+# Inc. All Rights Reserved.
+###############################################################################
+set -x -e
+
+# aptitude configuration
+APTITUDE_OPTIONS=${APTITUDE_OPTIONS:-"-y"}
```
spladug: perhaps these scripts could source a common config file rather than duplicating config like this?

spladug:
:no_bell: pet peeve: "setup" is a noun not a verb!

spladug:
:nail_care: i like splitting up the script, but i think there's still some work to reduce duplication and maintain ergonomics. awesome work getting it to build on travis, loving those green checkmarks!

install/install_cassandra.sh
KeyserSosa: Attempted to refactor away one in 230b1ed.  It adds a dependency between two of the scripts, though.  Let me know if that seems reasonable.

install/install_apt.sh
KeyserSosa: Done.  I wasn't sure of preferred style for this.  e20cca6.

Vagrantfile
KeyserSosa: On closer reading of the [Vagrant docs](https://docs.vagrantup.com/v2/multi-machine/) it looks like yes.  [Unfortunately they don't say as much until several sections after they say "Commands that only make sense to target a single machine, such as vagrant ssh, now *require* the name of the machine to control." which initially threw me.]

KeyserSosa:
:haircut: 

@spladug I think I addressed everything except 'setup' (though I at least fixed the comments that I added).  Do you want me to rename the scripts `set_up_*` or turn it into a suffix `_setup.sh`?

install/install_cassandra.sh
```diff
@@ -0,0 +1,49 @@
...
+sudo apt-get install $APTITUDE_OPTIONS cassandra=1.2.19
+
+# we don't want to upgrade to C* 2.0 yet, so we'll put it on hold
+apt-mark hold cassandra || true
+
+# cassandra doesn't auto-start after install
+sudo service cassandra start
+
+# check each port for connectivity
+echo "Waiting for cassandra to be available, see source for port meanings..."
```
bsimpson63: the part about "see source for port meanings" isn't necessary anymore

install/reddit.sh
```diff
@@ -0,0 +1,639 @@
...
+
+###############################################################################
+# Configure Services
+###############################################################################
+
+# Configure Cassandra
+$RUNDIR/setup_cassandra.sh
+
+# Configure PostgreSQL
+$RUNDIR/setup_postgres.sh $REDDIT_SRC/reddit
```
bsimpson63: don't need the path anymore

install/install_cassandra.sh
```diff
@@ -0,0 +1,49 @@
...
+RUNDIR=$(dirname $0)
+source $RUNDIR/install.cfg
+
+if [ ! -e $CASSANDRA_SOURCES_LIST ]; then
+    echo "Cassandra repo not added.  Running `install_apt.sh`"
+    $RUNDIR/install_apt.sh
+fi
+
+# grab the new ppas' package listings
+sudo apt-get update
```
bsimpson63: hasn't this already run?

bsimpson63:
:fish: lgtm

spladug:
:fish: looks great! thanks for all that

PR #1508, opened 2016-01-16
Update year range to 2016

pra85:


pra85:
@13steinj  I have updated the pull request so that there are no more merge conflicts. Also while browsering through the code I found some files ( like
https://github.com/reddit/reddit/blob/master/r2/r2/templates/printable.iframe and https://github.com/reddit/reddit/blob/master/r2/r2/templates/filteredinfobar.html ) which have older year license headers. Should those be updated as well ?

13steinj:
@pra85 if you seriously did this manually I'm both impressed and shocked at the same time. [I wrote a script which would do it automatically](https://github.com/reddit/reddit/commit/af09fa8dee69bef4f65a1662d9ad91c2329946e1#commitcomment-15379060)

13steinj:
Oh, I see what you mean now, seems @spladug missed some files back then. I'll just pr on your branch in 30ish minutes after I download the branch on my pc and run a better version of the script I wrote.

pra85:
@13steinj I have updated this pull request with your changes

13steinj:
Good; means they're all updated now. Just wait for an admin to review.

umbrae:
Thanks, we prefer to do this ourselves because we tend to catch all instances including all plugins, etc. This comes up every year. ;)

PR #1512, opened 2016-01-21
Testing: add vote unit test and coverage script

KeyserSosa:
This adds a rudimentary "votes work as expected" unit test as well as a script to test code coverage using `nosetests` and the `coverage` package which has been added to `setup.py`

:eyeglasses: @Deimos @spladug

### DIFF
--- a/r2/coverage.sh
+++ b/r2/coverage.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+set -e
+
+BASEDIR=$(dirname $0)
+cd $BASEDIR
+
+VERSION=$(git rev-parse HEAD)
+COVERDIR="/tmp/cover-$VERSION"
+
+function usage() {
+    echo "Run unit tests and coverage reports on reddit codebase with optional"
+    echo "http server to the report"
+    echo
+    echo "Usage: `basename $0` [options]";
+    echo
+    echo "  -h           show this message"
+    echo "  -p \$PORT     run an simple http server on \$PORT to view results"
+    echo "  -v           verbose mode (set -x)"
+    echo
+}
+
+while getopts ":vhp:" opt; do
+  case $opt in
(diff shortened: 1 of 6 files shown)


r2/setup.py
```diff
@@ -89,7 +89,7 @@
     # * https://pythonhosted.org/setuptools/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies
     # * https://github.com/pypa/sampleproject/blob/300f04dc44df51492deb859ac98ba521d2c7a17a/setup.py#L71-L77
     extras_require={
-        'test': ['mock', 'nose'],
+        'test': ['mock', 'nose', 'coverage'],
```
spladug: huh, what about putting these in `tests_require` instead of an extras?

r2/coverage.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+set -e
+
+BASEDIR=$(dirname $0)
+cd $BASEDIR
+
+VERSION=$(git rev-parse HEAD)
+COVERDIR="/tmp/cover-$VERSION"
```
spladug: how about putting this in `build/` so it doesn't rely on system paths?

KeyserSosa:
:haircut: 

@spladug `python setup.py test` now works with those `setup.py` changes but I had to make some changes to `app_globals.py` which was doing some (rather infuriating to track down) mocking based on a perceived "environment".

KeyserSosa:
:eyeglasses: @Deimos @spladug 

Apparently harold didn't pick up my ninja-edit of the reviewers list.  Being all explicit-like.

spladug:
:fish:

KeyserSosa:
merged in af49fa322bd018dbe69a89a75f4e22fbef6521b3

PR #1513, opened 2016-01-21
Update License Headers to 2016

lance386:


13steinj:
Duplicate of #1508 however it looks like you got more files changed. If I may recommend a double check, you can try running [this](https://github.com/reddit/reddit/commit/af09fa8dee69bef4f65a1662d9ad91c2329946e1#commitcomment-15379060)

umbrae:
Thanks, we prefer to do this ourselves because we tend to catch all instances including all plugins, etc. This comes up every year. ;)

PR #1520, opened 2016-01-31
Add data URI and font-face support to the CSS filter

novovnz:
The only factor stopping the font-face CSS at-rule from use is apparently that [it would have external references](https://www.reddit.com/r/cssnews/comments/24anzb/css_change_the_filter_has_been_rewritten/ch5n8b6). This is a solution to that which allows it to be used internally, with no extra infrastructure; I doubt it was intentionally disabled.

[Data URIs](https://css-tricks.com/data-uris/) are a way of encoding files in a CSS style. Any MIME file can be usually used, but the code in this request limits it to the existing image types and fonts (TTF). While a complete font-family is out of the question due to the 100kb limit, this would be quite useful for iconfonts, emoticons, and fancy heading text.

### DIFF
--- a/r2/r2/lib/cssfilter.py
+++ b/r2/r2/lib/cssfilter.py
@@ -398,17 +398,18 @@ class StylesheetValidator(object):
         self.images = images
 
     def validate_url(self, url_node):
-        m = SUBREDDIT_IMAGE_URL_PLACEHOLDER.match(url_node.value)
-        if not m:
-            return ValidationError(url_node.source_line, "NON_PLACEHOLDER_URL")
+        if not url_node.value.startswith(("'data:font/opentype", "'data:application/font-woff", "'data:image/jpg", "'data:image/png")):
+            m = SUBREDDIT_IMAGE_URL_PLACEHOLDER.match(url_node.value)
+            if not m:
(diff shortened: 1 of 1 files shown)


13steinj:
To weigh in: I'd love to be able to use this. Except for the fact that sometimes the amount of chars in a data URI would make the stylesheet go over the 100Kib limit, but I still like the idea (for smaller things as you mentioned).

However, there may be a problem. I don't work for reddit (nor do I know anything about security), but it seems that this would open up chances for [XSS vulnerabilities](http://www.gnucitizen.org/blog/bugs-in-the-browser-firefoxs-data-url-scheme-vulnerability/) (correct me if I'm wrong), which is why adding `filter` to the css filter was denied (older internet explorer versions had something with activex that was dangerous, I don't remember exactly what.)

novovnz:
That exploit can only be done in HTML; html and javascript URIs (even ones disguised as images/fonts) are not usable in CSS

Pokechu22:
Except for [expressions in IE7](https://msdn.microsoft.com/en-us/library/ms537634(v=vs.85).aspx) which actually allows arbitrary javascript in CSS (though that isn't allowed on reddit).

umbrae:
:eyeglasses: @spladug

bsimpson63:

PR #1521, opened 2016-01-31
Add community rule placeholders

13steinj:
As I suggested [here](https://www.reddit.com/r/AutoModerator/comments/42toej/feature_suggestion_shorthand_for_subreddit_rules/). I'm doing this entirely off mobile via reading past commits, so you might want to do a more thorough check than usual.

~~If you'd like I can also make it return an error to the mod that is editing the config (which IIRC is usually done for syntax errors) if a rule that they chose doesn't actually exist.~~ After looking into it, while definitely doable, I don't think that's a good thing to do, since you'd have to take into account the fact that rules can change, as well as the fact that the status of "when/howoften" is it checked becomes an issue. It seems safer to allow the mods to see  the fact that rules changed when they see a stray comment.


umbrae:
Thanks, we're already working on this one in-house. We'll see if we can use some of it for what we build.

13steinj:
@umbrae Any update on this? (Unless it's been added already and I haven't been keeping up with news).

Issue #153, opened 2011-07-12
Migrate from PyCAPTCHA to reCAPTCHA


spladug:
PyCAPTCHA hasn't been updated since 2006.

akkatracker:
@spladug am I right to assume that reCaptcha has been implemented everywhere and this issue can be closed?
![clipboard01](https://cloud.githubusercontent.com/assets/2032746/24345602/a9498f1a-131c-11e7-8fef-fc743dc9a5e0.jpg)

spladug:
Yup!

PR #1525, opened 2016-02-07
Update captcha.py

paragonie-scott:
PyCaptcha's `randomIdentifier()` uses `random.choice` which is a PRNG based on the Mersenne Twister. 

Let's use a CSPRNG instead. Otherwise this attack exists:

1. Get several (624) subsequent captcha images.
2. Get a human to translate them.
3. Crack the MT seed.
4. Bypass the CAPTCHA forever without human intervention.

### DIFF
--- a/r2/r2/lib/captcha.py
+++ b/r2/r2/lib/captcha.py
@@ -26,7 +26,6 @@ import random, string
 
 from pylons import app_globals as g
 
-from Captcha.Base import randomIdentifier
 from Captcha.Visual import Text, Backgrounds, Distortions, ImageCaptcha
 
 from r2.lib.cache import make_key
@@ -48,10 +47,10 @@ class RandCaptcha(ImageCaptcha):
                  Distortions.SineWarp()))
 
 def get_iden():
-    return self.randomIdentifier(length=IDEN_LENGTH)
+    return randomIdentifier(length=IDEN_LENGTH)
 
 def make_solution():
-    return self.randomIdentifier(alphabet=string.ascii_letters, length = SOL_LENGTH).upper()
+    return randomIdentifier(alphabet=string.ascii_letters, length = SOL_LENGTH).upper()
 
 def randomIdentifier(alphabet = string.ascii_letters + string.digits,
                      length = 24):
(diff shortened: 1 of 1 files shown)


garymoon:
Just out of curiosity, is there any longer a reaon not to switch to reCAPTCHA? As I understand it the reasons against mentioned in [this issue](https://github.com/reddit/reddit/issues/153) (several years old) no longer apply.

r2/r2/lib/captcha.py
```diff
@@ -48,10 +48,14 @@ def getLayers(self, solution="blah"):
                  Distortions.SineWarp()))
 
 def get_iden():
-    return randomIdentifier(length=IDEN_LENGTH)
+    return self.randomIdentifier(length=IDEN_LENGTH)
```
13steinj: There isn't really a "self" to reference here. It's not done from within the class.

r2/r2/lib/captcha.py
```diff
@@ -48,10 +48,14 @@ def getLayers(self, solution="blah"):
...
 
 def make_solution():
-    return randomIdentifier(alphabet=string.ascii_letters, length = SOL_LENGTH).upper()
+    return self.randomIdentifier(alphabet=string.ascii_letters, length = SOL_LENGTH).upper()
```
13steinj: Same as above

r2/r2/lib/captcha.py
```diff
@@ -48,10 +48,14 @@ def getLayers(self, solution="blah"):
...
+
+def randomIdentifier(alphabet = string.ascii_letters + string.digits,
```
13steinj: Since you're defining your own randomIdentifier anyway, the import at the top is no longer needed

paragonie-scott: Sorry, I don't do much Python on a day-to-day basis. Thanks for the feedback.

spladug:
Hi, thanks for the patch! In production, we have several hundred application servers each with dozens of worker processes. The cluster as a whole is handling tens of thousands of requests per second. The likelihood that you'd make 624 sequential requests to the same worker process in the whole cluster and that no one else would make a request to that worker in the meantime which happened to consume a random number is practically impossible. Given that the existing CAPTCHA is pretty easy to crack on a pure image processing level I don't think this is a reasonable attack vector. We're looking at replacing the CAPTCHA altogether for other reasons, so I'm not going to merge this at this time.

PR #1526, opened 2016-02-07
Update CONTRIBUTING.md

paragonie-scott:
https://adamcaudill.com/2015/11/19/responsible-disclosure-is-wrong/

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,7 +19,7 @@ look like a bug, but they are not.  If you're not sure, make a post in
 
 3. **DO NOT PUBLICLY REPORT SECURITY BUGS**.
 If you have found a bug that can in some way compromise the security of the
-site or its users, please exercise [responsible
+site or its users, please exercise [coordinated
 disclosure](http://www.reddit.com/wiki/whitehat) and e-mail
 security@reddit.com. We will respond to you as soon as possible. Please note
 that e-mails which do not pertain to security vulnerabilities in reddit's code


spladug:
Hi, while I agree with the point of that article I don't think it applies as clearly to a production website as it does to something like a traditional release model software vendor (e.g. an operating system distribution). For a production website, getting production patched before disclosure is in my opinion the only responsible path. Given that, I'd like to keep the current wording in this document. Thanks for the patch!

paragonie-scott:
> For a production website, getting production patched before disclosure is in my opinion the only responsible path.

Is this a correct interpretation of what you're saying?

* If it's a package that could potentially affect millions of computers, it's open to interpretation. Full disclosure *might* be okay.
* If it's one specific website, it's rigid and defined. Full disclosure *isn't* okay.

spladug:
As the post you linked mentioned, it's all about the users first. If there's a situation where broad disclosure before at least attempting to get the site owner to fix it in production first would somehow improve user security on a single site, I can't think of it.

paragonie-scott:
The scenario is this:

You inform users that they are at risk, and they leave the site and go to alternatives instead. This is bad for the website, but the users can still enact their agency and avoid being harmed once they are informed of the vulnerability already present in the website.

That's what full disclosure does to cryptography primitives, anyway.

spladug:
I think that's perfectly reasonable once it's been made clear that the site operators are not interested in fixing any issues. But by doing full disclosure up front you cause potential harm to the users before they get to "enact their agency" (say because they're on vacation and won't be back for a week or simply didn't see your post before the bad people did). If you disclose privately to the site operator first and they patch it, then the window of harm is negated.

paragonie-scott:
> I think that's perfectly reasonable once it's been made clear that the site operators are not interested in fixing any issues.

That's fair, but you said you couldn't think of any scenarios where full disclosure would help and/or be the more responsible thing to do, so I gave you one.

If you combine this with "the site operators are negligent and/or malicious" a'la @danielkerr of OpenCart, then it makes more sense to go to immediate full disclosure.

We're still talking hypotheticals.

PR #1528, opened 2016-02-08
Add queue counters for mods on the sidebar.

13steinj:
[As originally requested here](https://www.reddit.com/r/ideasfortheadmins/comments/43wcas/add_a_visible_counter_next_to_the_mod_queue_and/).

Also adds the basic framework to set actions to run client side every x amount of seconds, with x being configurable via interaction with the config.

The reasoning for not setting up an entirely different endpoint to get the counts is that it seemed unnecessary, the queues are pointless if they have more than 100 items anyway, and it would be more changes than necessary

E: it may be better to use the websocket framework to send reporting events and approve / remove events instead now that I think about it.


bsimpson63:

PR #1527, opened 2016-02-07
Add api endpoint to get saved subreddits

13steinj:
As requested [here](https://www.reddit.com/r/ideasfortheadmins/comments/44neh6/api_endpoint_to_get_saved_subreddits/)

PR #1529, opened 2016-02-09
Add a 'only' url parameter and menu for /u/me/saved for reddit gold users.

13steinj:
Allows gold users to filter /u/me/saved by type (link or comment).

Also adds a menu to do so.

And adds the only parameter to /api/saved_categories so that devs can make use of this as well. If this PR goes through, and so does #1527, #1527 should be done first so that the proper modifications to that endpoint can be made here as well.


bsimpson63:

PR #1533, opened 2016-02-21
Patch VSubredditName from experiencing a UnidcodeDecodeError

13steinj:
Resolve a potential `UnicodeDecodeError` in `VSubredditName`. Since names that would cause these errors aren't considered valid by the regex check in `Subreddit.is_valid_name`, and if a name is invalid it's not used in any action, it's safe to wipe them with a `""`. Originally reported [here](https://www.reddit.com/r/bugs/comments/46sxg5/that_500_error_when_trying_to_create_a_subreddit/)

E: As can be seen by [/r/13steinj_chars_test/](https://www.reddit.com/r/13steinj_chars_test/), there is no issue on any other fields.

### DIFF
--- a/r2/r2/lib/validator/validator.py
+++ b/r2/r2/lib/validator/validator.py
@@ -701,7 +701,10 @@ class VSubredditName(VRequired):
             name, allow_language_srs=self.allow_language_srs)
         if not valid_name:
             self.set_error(self._error, code=400)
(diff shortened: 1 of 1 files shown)


13steinj:
I'm closing this because @bsimpson63 's e2c5b95 has fixed the issue in essentially the same manner.

PR #1539, opened 2016-02-28
Precise bans

13steinj:
Allow mods to choose bans that are more precise, minimum of an hour.


umbrae:
Thanks for the effort, sadly I think day-long bans are sufficient.

PR #1542, opened 2016-03-04
reddit.less: listing chooser fix

Undust4able:
listing chooser sidebar now does have a fixed position and becomes sticky, so it does not scroll with the page.
its contents are intended instead so it looks prettier.
what's the purpose of having a sidebar that displays through the whole page but stays on top?

gif before: https://i.gyazo.com/077b2a731b2895925453237739d89269.gif
gif after: https://i.gyazo.com/a0b1d032616237bc62ab9d11d510c8d9.gif

### DIFF
--- a/r2/r2/public/static/css/reddit.less
+++ b/r2/r2/public/static/css/reddit.less
@@ -9746,8 +9746,8 @@ body.with-listing-chooser {
     }
 
     .listing-chooser {
-        position: absolute;
-        top: 65px;
+        position: fixed;
+        top: 0;
         left: 0;
         bottom: 0;
         width: @lc-width;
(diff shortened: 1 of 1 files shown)


13steinj:
Normally I'd comment on the reddit post but my mobile app isn't working atm. And I wanted to point something out. (in case your confused by my avatar, no, I'm not an admin).

I'm highly against this due to the implications. While the listing chooser itself is sticky, and the only problem I have with that is that I don't like the ux of that, there's a bigger problem of what's within the chooser.

Take a closer look at your gifs. Not only does the chooser icon stay in place, so does what is within the chooser. Now, that's fine if you only have a few things in there. But people with more multis wouldn't have quick and easy access to them.

bsimpson63:

PR #1545, opened 2016-03-05
Fixed deprecated form of raising exception

HeliumSquid:
Changed `raise NotFound, 'Account %s' % name` to `raise NotFound('Account %s' % name)`.

### DIFF
--- a/r2/r2/models/account.py
+++ b/r2/r2/models/account.py
@@ -339,7 +339,7 @@ class Account(Thing):
             return False
 
         return True
-    
+
     @classmethod
     @memoize('account._by_name')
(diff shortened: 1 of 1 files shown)


13steinj:
While I'd personally accept this kind of PR (don't get confused by my avatar, I'm not an admin):

1. This isn't the only time it's used like this.
2. Not really deprecated since reddit is Py2, and I don't see it switching to Py3 any time soon without a *huge* refactor (they'd have to switch to pyramid instead of pylons).
3. Read [rule 5 of CONTRIBUTING.md](https://github.com/reddit/reddit/blob/master/CONTRIBUTING.md)

PR #1547, opened 2016-03-09
Tried to make it attempt to make itself root if it's not root

Quartz101:
Not sure if it works, too lazy to try right now.


kjoconnor:
This doesn't work. Please test it before reopening.

PR #1552, opened 2016-03-20
Persistent reports

13steinj:
This will allow mods to view the reports of an item even after an action has been taken. A side effect is a slight behavior change in the ignoring of reports for an item, which is mentioned in detail within the commits.


r2/r2/public/static/css/reddit.less
```diff
@@ -4498,6 +4498,9 @@ form input[type=radio] {margin: 2px .5em 0 0; }
     border: 1px solid black !important;
     padding: 0 4px;
     background-color: #f6e69f;
+    &.new-reports {
+        background-color: #ff0000;
```
13steinj: I chose pure red for lack of a better idea.

empyrical:
How hard would it be to have relevant mod actions punctuate the reports? like:

```
1: <no reason>
 * MOD approved X hours ago
1: breaks rule #2
 * MOD removed X hours ago
```

13steinj:
@empyrical doable? definitely. Reasonable, that depends on how tough the caching is which I can't find out properly. And since that data becomes overwrriten by redoing an action, that means that the data  "model" would have to be changed completely, unless you're okay with potentially another who knows how long backfill.

13steinj:
I rebased and force pushed so that Travis will pass (it was broken because of an error in LOID cookies, which was broken earlier and fixed as of #1553 's merge).

bsimpson63:

PR #1554, opened 2016-03-22
VCommentIDs: remove duplicate ids

13steinj:
GET_morechildren is exploding down the line if there's duplicates in the id list. The simple solution is to split via regex to account for spaces, and use `set` on it to remove duplicates.

### DIFF
--- a/r2/r2/controllers/api.py
+++ b/r2/r2/controllers/api.py
@@ -3561,6 +3561,7 @@ class ApiController(RedditController):
                 return abort(403,'forbidden')
 
             if children:
+                children = list(set(children))
                 builder = CommentBuilder(link, CommentSortMenu.operator(sort),
                                          children=children,
                                          num=CHILD_FETCH_COUNT)


umbrae:
:eyeglasses: @bsimpson63

bsimpson63:
What does it mean that it's exploding?

13steinj:
@bsimpson63 if duplicate ids are added a [500 internal server error pops up](http://www.reddit.com/r/bugs/comments/4bfxx2/api_getting_http_status500_when_stupidly_supplying_duplicate_ids_to_the_morechildren_endpoint/). Haven't had the time to boot up my local install and get the log, but I can get it today

bsimpson63:
The exception is happening here https://github.com/reddit/reddit/blob/master/r2/r2/controllers/api.py#L3587

`AttributeError: 'NoneType' object has no attribute 'things'`

I think it'd be better to fix that (by reducing `comments` to the unique set before passing into `CommentBuilder` or figuring out what's going wrong within `CommentBuilder`) rather than mess with the validator.

bsimpson63:
:nail_care:

13steinj:
:haircut: @bsimpson63

bsimpson63:
:fish: thanks

bsimpson63:
118106e5fbbd0f4f07f6bc357954ebed8d3e4ad4

PR #1555, opened 2016-03-26
Make the "get started with automod" go to the sr wiki page if there's…

13steinj:
… no docs

The button as it stands shows up on local installs, even if automod is not enabled. The button also leads to /wiki/automoderator, even though such a page with docs doesn't necessarily exist.

This makes the button only show up if automod is enabled on the install, and if there are no docs, it sends the button to the "create config/automoderator" page instead of a 404.

### DIFF
--- a/r2/r2/lib/pages/pages.py
+++ b/r2/r2/lib/pages/pages.py
@@ -589,8 +589,8 @@ class Reddit(Templated):
                     css_class="reddit-flair access-required",
                     data=data_attrs('flair')))
 
-        if is_single_subreddit and is_moderator_with_perms('config'):
-            # append automod button if they have an AutoMod configuration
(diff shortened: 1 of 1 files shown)


Deimos:
I'm okay with the first part of this (not showing the button if `g.automoderator_account` isn't set), but I don't like the other change. On the production site, that's going to add another wiki page lookup on every page load to check if the docs (still) exist, for no benefit at all.

If you just want to cut it down to the first change that's probably fine. If you do that, make sure to shorten the lines as well, they're way over the length limit.

:nail_care:

13steinj:
:haircut: @Deimos

Deimos:
:fish:

Deimos:
https://github.com/reddit/reddit/commit/117d7f06d6ff5f2f9690f423ee2bf595a4be08dc

PR #1556, opened 2016-03-26
Add author flair placeholders to automoderator

13steinj:
As requested and reminded [here](https://www.reddit.com/r/ModSupport/comments/4bz7xg/an_icon_for_1_day_old_accounts/d1e4f1k), adds placeholders for the author's flair css class and text.

Defaults to an empty string if the attribute is missing (they don't a flair).

### DIFF
--- a/r2/r2/lib/automoderator.py
+++ b/r2/r2/lib/automoderator.py
@@ -98,6 +98,10 @@ def replace_placeholders(string, data, matches):
         "{{author}}": data["author"].name,
         "{{body}}": getattr(item, "body", ""),
         "{{subreddit}}": data["subreddit"].name,
+        "{{author_flair_text}}": data["author"].flair_text(
+            data["subreddit"]._id, obey_disabled=True),
+        "{{author_flair_css_class}}": data["author"].flair_css_class(
+            data["subreddit"]._id, obey_disabled=True),
     }
 
     if isinstance(item, Comment):


r2/r2/lib/automoderator.py
```diff
@@ -98,6 +98,10 @@ def replace_placeholders(string, data, matches):
         "{{author}}": data["author"].name,
         "{{body}}": getattr(item, "body", ""),
         "{{subreddit}}": data["subreddit"].name,
+        "{{author_flair_text}}": getattr(data["author"],
```
Deimos: This isn't quite the right way to get their flair, since it won't take into account whether they've disabled their flair in the subreddit. I would put something like this before the `replacements = {` line above:

```
# get author flair text/class
sr_id = data["subreddit"]._id
author_flair_text = data["author"].flair_text(sr_id, obey_disabled=True)
author_flair_css_class = data["author"].flair_css_class(
    sr_id, obey_disabled=True)
```

And then here, just do (note that I'm also changing `{{author_flair_class}}` to `{{author_flair_css_class}}` for consistency with the names AutoMod uses for flair checks):

    "{{author_flair_text}}": author_flair_text,
    "{{author_flair_css_class}}": author_flair_css_class,

It's not necessary to worry about the case where the user doesn't have flair (or if they've disabled it in that subreddit), since `_force_unicode(None)` returns an empty string anyway, so it'll already work properly.

Deimos:
:nail_care:

r2/r2/lib/automoderator.py
Deimos: Actually doing the section separately the top is probably unnecessary, I changed my mind about doing some other stuff in there so it's kind of pointless now. Might as well just do it right in here:

```
            "{{author_flair_text}}": data["author"].flair_text(
                data["subreddit"]._id, obey_disabled=True),
            "{{author_flair_css_class}}": data["author"].flair_css_class(
                data["subreddit"]._id, obey_disabled=True),
```

13steinj:
:haircut: @Deimos

Deimos:
:fish:

Deimos:
https://github.com/reddit/reddit/commit/b04c6f42b91c5c91d58a88685831adbbaf1991f1

PR #1565, opened 2016-04-08
mobile: Fix link issue in message about gold.

yowar:
reddit
The new mobile applications do not handle relative URLs.


dellis23:
Thanks!

PR #1568, opened 2016-04-09
Expose GET_comment_by_id to non admins

13steinj:
This will allow users to get to a comment via /c/commentid, which could be useful.

Since the same logic to hide private sub's applies as it were in the link id redirect, this also factors that out into a separate function to reduce on repitition.

### DIFF
--- a/r2/r2/controllers/front.py
+++ b/r2/r2/controllers/front.py
@@ -76,15 +76,22 @@ class FrontController(RedditController):
 
     allow_stylesheets = True
 
-    @validate(link=VLink('link_id'))
-    def GET_link_id_redirect(self, link):
-        if not link:
+    def _make_thing_redirect(self, thing, **kwargs):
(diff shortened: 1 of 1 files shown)


empyrical:
would also open up the thing i attempted in #1063

13steinj:
@empyrical didn't know! sorry to steal the thunder (if that's what I did).

I don't personally agree on the comment shortlinks though, since that's not really something that can be changed from an outsider (in production, they use a specific short domain and they'd need to change that on their end as well)

empyrical:
nah no worries,i closed it because it became unmergable and it needed to be rewritten anyways (its almost 2 years old now lol)

i spoke to them on irc when i made it and they said that they just need to change something in their load balancer to allow `redd.it/c/xxxxx` to work

bsimpson63:

PR #1569, opened 2016-04-09
install-reddit.sh: Fetch the install scripts without root.

13steinj:
install-reddit.sh currently fetches the individual install scripts as root. Since the person installing reddit may be inept at linux, they may not be able to figure out how to edit these files on the chance they need to (ex editing install/install.cfg to add another plugin).

This change makes the directory be made and the files be retrieved as the user who had started the script, so they won't have an annoyance editing the retrieved files later on.


bsimpson63:

PR #1572, opened 2016-04-10
mobile: Fix link issue in message about gold.

yowar:
The new mobile applications do not handle relative URLs.
photo


yowar:
photo0135

PR #1582, opened 2016-04-16
Fixed parentheses in setup.py

xshotD:
Line 36's print statement didn't contain the required parentheses. Always keep attention to your syntax!


spladug:
Thanks for your contribution. However, this is a Python 2.7 project and `from __future__ import print_function` is not in effect. So these are print statements, not calls to the print function. No parentheses are necessary for print statements.

PR #1583, opened 2016-04-17
set_api_docs: skip VNotInTimeout validators without docs

13steinj:
VNotInTimeout is causing the details of some parameters to be wiped, such as the details of [id in POST_set_subreddit_sticky](https://www.reddit.com/dev/api#POST_api_set_subreddit_sticky), because by default validator docs are `{param: None for param in self.param}`. If VNotInTimeout is being called within the `@validate` or `@validatedForm` decorators, and is being called with a parameter (ex 'id'), the docs for that parameter get replaced with `None`.

This solves the issue by skipping VNotInTimeOut if there are no specified docs.


bsimpson63:

PR #1585, opened 2016-04-20
Don't create the subreddit as gold only if the user isn't allowed to.

13steinj:
Currently the error is set but not checked for during creation, which if this is the only inconsistency, the subreddit would be made without fail. Found by [this post](http://www.reddit.com/r/bugs/comments/4fjop1/_/).


MelissaCole:
:fish: Thanks!

MelissaCole:
https://github.com/reddit/reddit/commit/f4b335a7289072d2b9152a242692d38f79f9ebf9

PR #1587, opened 2016-04-24
Box was 404

tienne-B:
The vagrant box gave a 404 error. This commit changes the trusty box to the most current (at the time of authoring)


lan17:
Please update to an exact version, not current.

kjoconnor:
We've decided to just always go with current absent a better solution. Thanks anyway!

https://github.com/reddit/reddit/commit/b1550aed10fd3ac7695c1ecb7ed15e308cdcb7ba

PR #1588, opened 2016-04-24
Add missing alt, height and width attribute to img

parlet:
Image of a pixel to pick up HSTS policies from the base domain don't have an alt attribute. HTML standard specifies that images not intended for the user (like this one) **must** have an alt attribute with empty value. It also mentions that "in such cases, the width and height attributes **should** both be set to zero". Based on [section 4.8.4.1.12 of HTML standard](https://html.spec.whatwg.org/multipage/embedded-content.html#an-image-not-intended-for-the-user).


bsimpson63:

PR #1589, opened 2016-04-24
Remove unused, invalid and nonsensual attribute

parlet:
\<div class="footer rounded"> have "by-zero" attribute. This is its only occurrence in the Reddit code, so it doesn't seem to serve any purpose, nor it is valid. It was added with commit "[oh god how did this get here I am not good with computer](https://github.com/reddit/reddit/commit/e175e5e1321e30edf69bddec1a898e62ad899fb2)".

### DIFF
--- a/r2/r2/templates/redditfooter.html
+++ b/r2/r2/templates/redditfooter.html
@@ -27,7 +27,7 @@
 <%namespace file="utils.html" import="text_with_links"/>
 
 <div class="footer-parent">
-  <div by-zero class="footer rounded">
+  <div class="footer rounded">
       %for toolbar in thing.nav:
       <div class="col">
         ${toolbar}


erikdesjardins:
It's a joke: "<div[iding] by-zero"

13steinj:
Hello captain buzzkill :P 

While yes completely nonsensical, just at a personal level, having fun is a good thing ;)

parlet:
@13steinj: Oh, now I get the intended pun (thanks @erikdesjardins), but man... that joke is so bad. And however good it wouldn't be, it shouldn't be pushed to production in a manner invalidating the code, if at all.

amit3vr:
I actually find it quite amusing :p

spladug:
Hi there, thanks for your contribution. It is indeed a joke and it's not really harming anyone so we're gonna keep it.

parlet:
Well, I don't think I could agree with that, it is invalid attribute that validator complains about. You could perhaps instead add a comment above this div that would read something along the lines "beware, below div is by zero" , if you really want to retain this joke.

PR #1599, opened 2016-05-05
Lossslessly compressed PNG images using PNGcrusher.

robotortoise:
I losslessly compressed the PNG images using PNGcrusher.

It reduced the file size down quite a bit. Maybe I'll make some of the
simpler PNG images vectors in the future.


robotortoise:
...What?

How did the build fail?! I just changed the PNGs, right...?

Pokechu22:
That probably isn't your fault.  It looks like it's [failing](https://travis-ci.org/reddit/reddit/builds/127858996) after 3ea40a0.

robotortoise:
Okay, thank you.

So what should I do, then? Just...take the files from before that branch and upload to a new fork?

PR #1600, opened 2016-05-05
Bug fixes and improvements to proxy support

strix-technica:
1. Make IPs trusted for evaluating `X-Forwarded-For` configurable with new INI keyword `trusted_proxy_ranges`, a comma-separated list of CIDR specs (or bare IPs).  Move hard-coded trusted IPs 127.0.0.1 or 10.0.0.0/8 into example.ini.
2. Bug fixes in `X-Forwarded-For` processing:
   - remove whitespace around extracted IP which causes the
      memcached exception 'bad key' (see [1] for an example
      stacktrace)
   - extract correct IP when multiple IPs are present.  The de
      facto spec[2] says that each proxy _appends_ its peer's
      IP, which means the original requestor is the first IP,
      not the last.

[1] http://pastebin.com/1QQiRsgs

In this instance, the request is https terminated by nginx and then
routed to Reddit via Varnish cache (instead of HAProxy) which
automatically appends the remote IP per the de facto spec[2] for the
X-Forwarded-For header.

Because there are effectively two proxies involved, the XFF header
contains two IP addresses separated by a comma and a space.  Having
instrumented lib/cache.py:634, argument `keys` contained (note the
leading whitespace) `[' 127.0.0.1']`.  See the stack trace for the
original contents of the XFF header.

[2] X-Forwarded-For is an uncodifed, de facto standard (now replaced
    by RFC 7239), however there are a number of reputable resources
    that show how it is supposed to work, for example:

http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/x-forwarded-headers.html#x-forwarded-for
https://www.citrix.com/blogs/2012/09/27/multiple-xff-ips/


strix-technica:
Per [Pokechu22's comment](https://github.com/reddit/reddit/pull/1599#issuecomment-217071729), the travis failure is apparently unrelated to this or PR #1599.

spladug:
Hi, thanks for your contribution!

Patching this function is totally something we'd accept. However, I think it'd be cleaner / more flexible to make it so that you read a CIDR range (or list of CIDR ranges) from the config and just check against that range here:

    def is_trusted_proxy(ip):
         for range in g.trusted_proxies:
            if ip in range:
                return True
         return False

in dev, example.ini might look like

    trusted_proxies = 127.0.0.1/32

while in prod it might look like

    trusted_proxies = 10.0.0.0/8

There's an ipaddress library (backported from py3) in r2.lib.contrib which can help with this.

r2/r2/lib/base.py
```diff
@@ -104,7 +107,7 @@ def __call__(self, environ, start_response):
             request.ip = cdn_ip
             request.via_cdn = True
         elif g.trust_local_proxies and forwarded_for and is_local_address(remote_addr):
-            request.ip = forwarded_for.split(',')[-1]
+            request.ip = forwarded_for.split(',')[0].strip()
```
spladug: This reads from the opposite end of the list of forwarded-fors, is that intended?

r2/example.ini
```diff
@@ -150,6 +150,8 @@ static_domain =
 # if your webserver is a proxy and on a different instance on the same 10.0.0.0/8 network
 # set X-forwarded-for and set this to true
 trust_local_proxies = false
+# List of trusted proxy IPs.  Only neded if your proxies aren't on localhost or 10.0.0.0/8
```
spladug: > neded

spladug:
:nail_care:

strix-technica:
> Patching this function is totally something we'd accept. However, I think it'd be cleaner / more flexible to make it so that you read a CIDR range (or list of CIDR ranges) from the config and just check against that range here:

Great!  

And a CIDR spec list is an excellent idea which had not occurred to me.  I'll make the necessary adaptations and update the PR in the next day or two.

Please see above in-line note regarding the question of the order of XFF IPs.

spladug:
> Please see above in-line note regarding the question of the order of XFF IPs.

apologies, I will learn to read. :)

r2/r2/lib/base.py
strix-technica: > This reads from the opposite end of the list of forwarded-fors, is that intended?

It is.  According to:

http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/x-forwarded-headers.html#x-forwarded-for and
https://www.citrix.com/blogs/2012/09/27/multiple-xff-ips/

the X-Forwarded-For header takes the form:

> X-Forwarded-For: OriginatingClientIPAddress, proxy1-IPAddress, proxy2-IPAddress

This is further corroborated by the behaviour of Varnish cache wrt the XFF header.

Since `request.ip` is supposed to be the requesting client's IP, not that of any of the intermediate proxies, the line in question should be taking the first IP in the list, not the last.

With standard HAProxy installations, this bug would probably never surface because there would never be more than one IP in XFF.  But, in my Varnish-based install, SSL connections are first terminated by nginx, which proxies to Varnish, which proxies on to the paste reddit process so XFF contains two IPs and, as currently written, reddit records the wrong IP.

God knows how the business end of production reddit goes in respect of proxies, caches and load-balancers.  As a computer scientist and electronic engineer by profession, I'd love to know, but I guess that probably comes under the heading of 'trade secret'.

Without the strip() call, [errant whitespace causes memcached exceptions](http://pastebin.com/1QQiRsgs).  Per the PR text, I instrumented the code (with a little guidance from bsimpson) and the culprit key was (note the leading whitespace in the string):
`[' 127.0.0.1']`

strix-technica:
> apologies, I will learn to read. :)

Oh, you were too quick for me!  I hadn't yet written the comment to which I'd referred :)

But n/m.  I guess there aren't many doubt left now ;)

strix-technica:
:haircut: 

@spladug CIDR support now pushed.  Thanks for the tip re. r2.lib.contrib.ipaddress!

strix-technica:
I didn't intend to include 7659bcc in this PR, but it managed to find its way in there anyway.

All it does is prevent an exception when sending emails so it won't do any harm to be included upstream.  If there is a better or more complete solution to that particular bug, I guess it'll be addressed in its own good time.

r2/r2/lib/base.py
```diff
@@ -43,12 +44,17 @@
...
-    else:
-        trusted_ips = []
-    return ip.startswith('10.') or ip == "127.0.0.1" or ip in trusted_ips
+def is_trusted_proxy(ip):
+    ip = ipaddress.IPv4Address(ip)
+    trusted_ips = getattr(g, 'trusted_proxy_ips', '').strip()
+    if not trusted_ips:
+        return False
+    trusted_ips = [x.strip() for x in trusted_ips.split(',')]
+    trusted_ips = [ipaddress.IPv4Network(x.strip(), False) for x in trusted_ips]
```
spladug: This can be kinda expensive to do, it'd be better to pre-build this list of objects in `app_globals` so it's not re-running on every request.

r2/r2/lib/base.py
```diff
@@ -43,12 +44,17 @@
...
 
-def is_local_address(ip):
-    if hasattr(g, 'trusted_proxy_ips'):
-        trusted_ips = [x.strip() for x in g.trusted_proxy_ips.split(',')]
-    else:
-        trusted_ips = []
-    return ip.startswith('10.') or ip == "127.0.0.1" or ip in trusted_ips
+def is_trusted_proxy(ip):
+    ip = ipaddress.IPv4Address(ip)
+    trusted_ips = getattr(g, 'trusted_proxy_ips', '').strip()
```
spladug: This parsing in here is unnecessary if you use the config parsing layers, in `app_globals` you could add a section of `ConfigValue.tuple_of(ConfigValue.baseplate(ipaddress.IPv4Network))` (the baseplate thing is just to adapt the constructor of ipv4network to the expected protocol of those helpers) with `trusted_proxy_ips` (which should be renamed to `trusted_proxy_ranges` perhaps) and it'd handle parsing the comma-delimited line and converting to IPv4Network objects up-front rather than repeating it every request which is kinda messy and expensive (the ipaddress library is pretty slow).

r2/r2/lib/base.py
```diff
@@ -103,8 +112,8 @@ def __call__(self, environ, start_response):
         if cdn_ip:
             request.ip = cdn_ip
             request.via_cdn = True
-        elif g.trust_local_proxies and forwarded_for and is_local_address(remote_addr):
-            request.ip = forwarded_for.split(',')[-1]
+        elif g.trust_local_proxies and forwarded_for and is_trusted_proxy(remote_addr):
+            request.ip = forwarded_for.split(',')[0].strip()
```
spladug: > extract correct IP when multiple IPs are present. The de facto spec[2] says that each proxy appends its peer's IP, which means the original requestor is the first IP, not the last.

The reason we pull the last IP is that it's the only one we can trust, which is the point of checking `is_trusted_proxy`. Anyone can put whatever they want in the XFF headers that travel over the internet.

spladug:
:nail_care: for config parsing and XFF. thanks for working on this!

r2/r2/lib/base.py
strix-technica: I think there's some confusion here or we may be talking at cross purposes.

> The reason we pull the last IP is that it's the only one we can trust

As you say, XFF header can't be trusted unless the immediate peer (`environ['REMOTE_ADDR]`) is trusted (and, implicitly, this instance trusts the chain of IPs that the peer trusts etc).  

The new INI kw `trusted_proxy_ips` provides the list of trusted immediate peers.  I only check the immediate peer on the assumption that the trusted peer itself either replaces the XFF header entirely or trusts an upstream XFF header if present, but I suppose I could adapt `is_trusted_proxy()` to check all the IPs in the XFF chain if desired.  It seems superfluous, though, and would incur additional load.

Anyway, once the trustworthiness of the peer has been established, the XFF header can be considered valid.  At this point, the objective is to assign to `request.ip` the IP of the original requestor (just as original line 109 uses the peer's IP when the peer is not trusted or there's no XFF header).

If XFF contains only one IP (as would be the case in default dev installs), both the first and last IP is that of the original requestor.  If, however, there is more than one IP in the XFF header (such as happens when Varnish cache is used in lieu of HAProxy), the first IP is the original requestor and every subsequent IP will be the IPs of machines in the trust chain with the last IP being the most recent proxy (and so should match `environ['REMOTE_ADDR]` if it is appending to the XFF as it should).

That's why the correct thing to do is to take the first IP rather than any of the subsequent IPs.

r2/r2/lib/base.py
strix-technica: Will make these changes and push in the next day or so.

strix-technica:
:haircut:

@spladug Requested changes are now pushed.

> and it'd handle parsing the comma-delimited line and converting to IPv4Network objects

Excellent!

strix-technica:
:haircut: 

Merge conflict resolved.

spladug:
Hey, I'm really sorry for the delay here. I've done some testing with our load balancers and don't think we should can do the swap to the first IP in XFF right now. Our HAProxy instances are just appending to the existing XFF header, so only the last IP in it can be trusted. I'm curious, are you using multiple proxies in a chain internal to your network? or is something else causing more than one trusted IP in your header?

spladug:
:nail_care:

bsimpson63:

PR #1603, opened 2016-05-12
Fix compiled CSS to cope with parent-relative paths

### LINKED ISSUE #1550: reddit.less image URLs are compiled incorrectly
Problem
=======

The `url` from the following LESS snippet (and others similar to it):
```less
.locked-infobar:before {
  @1x: url('../infobar-icon-lock.png');
```

compiles into the same `url` with `..`:
```css
.locked-infobar:before{background-image:url('../infobar-icon-lock.png');
```

When the resulting CSS is loaded **from a non-root path** e.g. https://www.reddit.com/static/reddit.css that rule points to a non-existent file https://www.reddit.com/infobar-icon-lock.png

**NOTE:** This doesn't affect reddit.com directly since it loads CSS files from the root path (https://www.redditstatic.com/reddit.css) where `url(../)` is luckily ignored, but still the resulting CSS is invalid and this is clearly a bug (and it is reproduced on a fresh reddit install with https://github.com/reddit/reddit/wiki/reddit-install-script-for-Ubuntu).

Expected
========

The `../` part of the image URL is removed from the resulting CSS and it points to correct location, e.g. https://www.reddit.com/static/infobar-icon-lock.png

strix-technica:
Compiled (static) CSS and sprites are now written out to subdirectory `static/compiled-css/` rather than `static/` so that parent-relative `url()` references are valid when `uncompressedJS = false`.  Also patches `static.py` to cope with path elements that weren't there before.  No changes to `.less` sources are required, so functionality with `uncompressedJS = true` remains unchanged.

Without this patch, any parent-relative assets (such as `url(../sidebar-grippy-hide.png)`) return 404 or, sometimes, 500 (see [example stacktrace](http://pastebin.com/Y6LDgjNA)) when `uncompressedJS = false` because relative URLs in the sources are not adjusted to compensate for the relocated compiled stylesheet output.

Resolves #1550 which describes the problem in more detail; implements proposed solution described in [this /r/redditdev post](https://redd.it/4i6tig) (which contains yet more detail).

### DIFF
--- a/r2/Makefile
+++ b/r2/Makefile
@@ -96,6 +96,7 @@ STATIC_FILES := $(shell find $(STATIC_ROOT) -readable)
 STATIC_BUILD_ROOT := $(BUILD_DIR)/public
 STATIC_BUILD_DIR := $(STATIC_BUILD_ROOT)/static
 STATIC_BUILDSTAMP := $(BUILD_DIR)/static-buildstamp
+CSS_DEST_DIR := $(STATIC_BUILD_DIR)/compiled-css
 
 .PHONY: clean_static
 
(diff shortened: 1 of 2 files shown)


bsimpson63:

PR #1604, opened 2016-05-13
Compressed PNG images using PNGcrusher. Made some larger images transparent and autocropped 'em.

robotortoise:
Compressed PNG images using PNGcrusher. Made some larger images
transparent and autocropped 'em.


Harry-Torry:
Voting this up to be merged.

umbrae:
Thanks for your efforts, but all sprited pngs are optimized during make time, which is most of them. See nymph.py and `nymph` in the Makefile for more info.

PR #1606, opened 2016-05-24
Correct assignment of users' flairtext and flair_css_class so that subreddit settings are taken into account.

Keerikkadan91:
Reddit continues to serve the author_flair_text field in the JSON even when the user flair has been disabled by the mods for a particular subreddit. This is because the flair_text is currently populated only based on users' preferences, ignoring the subreddit setting. A new argument has been introduced to fix this; a default value is specified such that the behaviour remains the same as was before this change, should the argument not be used.

Originally reported [here](https://redd.it/4kpzby).

### DIFF
--- a/r2/r2/lib/automoderator.py
+++ b/r2/r2/lib/automoderator.py
@@ -99,9 +99,9 @@ def replace_placeholders(string, data, matches):
         "{{body}}": getattr(item, "body", ""),
         "{{subreddit}}": data["subreddit"].name,
         "{{author_flair_text}}": data["author"].flair_text(
-            data["subreddit"]._id, obey_disabled=True),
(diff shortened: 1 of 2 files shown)


13steinj:
Just as a "code grammar" gripe -- never pass on a keyword argument if that argument is redundant from the outside. Meaning; `flair_enabled_by_sr` is a kwarg determined by an external factor.

Also, that's for automoderator. Which may be good to check flairs regardless of mod disabling, since mods could use this then as a hidden way to keep track of problem posters if flairs are disabled (mods can still manually set invisible userflairs).

The way it should have been done is in r2/lib/jsontemplates.py, where `if attr=='author_flair_text'` (/css), add a `if not thing.subreddit.flair_enabled` check.

bsimpson63:

PR #1612, opened 2016-06-01
Fix issue with moderator removal message

Jakeable:
When a moderator removes themself, a message would still be sent ([example](https://www.reddit.com/message/messages/5n38ss)). This stops that from happening.

Perhaps a better fix would be to send a slightly different message to the subreddit if a moderator removes themself (e.g. /u/user has removed themself as a moderator of /r/subreddit).


Deimos:
:eyeglasses: @KeyserSosa

KeyserSosa:
:fish: lgtm

KeyserSosa:
merged 36fc55578c0c5b7cb7e443aa7a9dd3a7cc9f0689

PR #1618, opened 2016-06-12
Allow users to select a ?only type on /gilded

13steinj:
This allows users to use the ?only parameter like they can on /about/ spamlistings to choose between gilded links, comments, or both, on /gilded. /comments/gilded remains for the sake of compatibility.


bsimpson63:

PR #1620, opened 2016-06-15
Allow users to view their own content regardless of can_view results

cjm-0b10000100111010:
Fixes #629

### DIFF
--- a/r2/r2/models/builder.py
+++ b/r2/r2/models/builder.py
@@ -364,8 +364,9 @@ class Builder(object):
         if hasattr(item, 'can_view_slow') and not item.was_comment:
             return not item.can_view_slow()
 
-        if hasattr(item, 'subreddit') and not item.subreddit.can_view(user):
-            return True
+        if (hasattr(item, 'subreddit') and not item.subreddit.can_view(user) and
+               (item.author is None or item.author is not c.user)):
+	    return True
 
     def _is_controversial(self, wrapped):
         """Determine if an item meets all criteria to display as controversial."""


13steinj:
Iirc this should work for comments, but does it for submissions as well? I mean one would be able to view (and edit?) comments via their userpage but last I checked the submission text is retrieved on the spot (and uneditable from the userpage).

colin-merc-tech:
> Iirc this should work for comments, but does it for submissions as well? I mean one would be able to view (and edit?) comments via their userpage but last I checked the submission text is retrieved on the spot (and uneditable from the userpage).

Submissions will appear in the userpage listing, but you are correct that editing them wouldn't be possible. From a UX standpoint, I think it makes sense that a user should be able to see a reference to their submitted content on their userpage listing, regardless of being able to edit the content. With that said, the direction I took with https://github.com/reddit/reddit/pull/1620/commits/950a8b89aec5114bc6917019ea41e6863b313830 was entirely based on the wording of  #629: 
> should allow users to view their own content

cjm-0b10000100111010:
> Iirc this should work for comments, but does it for submissions as well? I mean one would be able to view (and edit?) comments via their userpage but last I checked the submission text is retrieved on the spot (and uneditable from the userpage).

Submissions will appear in the userpage listing, but you are correct that editing them wouldn't be possible. From a UX standpoint, I think it makes sense that a user should be able to see a reference to their submitted content on their userpage listing, regardless of being able to edit the content. With that said, the direction I took with 950a8b8 was entirely based on the wording of #629:

>should allow users to view their own content

13steinj:
Yes, but the underlying post of that issue references editing and deleting as well (deleting can be done, but the submission can't be edited). Making editing possible should be as simple as replacing [this line](https://github.com/reddit/reddit/blob/master/r2/r2/controllers/front.py#L271) with:

    if c.user != article.author_slow:
        abort(403, 'forbidden')
    c.hide_comments = True
    c.site = DefaultSR  # hide sr sidebar and info

And then check for c.hide_comments before [adding the comments listing](https://github.com/reddit/reddit/blob/master/r2/r2/controllers/front.py#L435).

bsimpson63:

PR #1621, opened 2016-06-15
JS parens/brace fix

strix-technica:
Fix broken JS syntax from commit 0c6546d58 which caused `uglyfyjs` to break `make static`.  

Also fix indentation.

@birakattack, you might want to double check that my edit is consistent with your intentions.

### DIFF
--- a/r2/r2/public/static/js/reddit.js
+++ b/r2/r2/public/static/js/reddit.js
@@ -763,20 +763,18 @@ function reply(elem) {
         }
     }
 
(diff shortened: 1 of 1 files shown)


birakattack:
:fish:

strix-technica:
Cheers, @birakattack

birakattack:
thanks for the fix 👍

birakattack:
merged f56e810

PR #1624, opened 2016-06-17
Make mod removal message clear what user was removed

SmithsonianDSP:
Specify the user that was removed in the subject line (instead of `You've been removed...`, it's `{user} has been removed...`) and start the message body with `{user}: You have been been removed...` (instead of just `You have been removed`).

[Relevant reddit comment explaining reasoning for this change here](https://www.reddit.com/r/modnews/comments/4ly8br/demodding_now_messages_the_user_who_was_removed/d4d7c45):

>The way it currently is, you have to look quite closely at the to/from line in order to see who was removed (and by whom), and it's already caused a bit of confusion with people seeing the message in the modmail and thinking it was sent to them (which, I imagine, is not going to be uncommon; a lot of people might the subject "**You've been removed as a moderator of..**" and, confused/panicked, skip over the comparably smaller sized to/from section of the message).

>The phrasing of the message makes sense if you're the one **getting** it, but the removal message is also sent through modmail so everyone in the sub sees it in their modmail inbox and gets the new modmail notification. Maybe change the subject to "/u/removed_mod was removed as a moderator of /r/subreddit"?

### DIFF
--- a/r2/r2/lib/system_messages.py
+++ b/r2/r2/lib/system_messages.py
@@ -112,9 +112,9 @@ def notify_user_added(rel_type, author, user, target):
 
 def send_mod_removal_message(subreddit, mod, user):
     sr_name = "/r/" + subreddit.name
-    subject = "You've been removed as a moderator from %(subreddit)s"
+    subject = "%(user)s has been removed as a moderator from %(subreddit)s"
     message = (
-        "You have been removed as a moderator from %(subreddit)s.  "
+        "%(user)s: You have been removed as a moderator from %(subreddit)s.  "
         "If you have a question regarding your removal, you can "
         "contact the moderator team for %(subreddit)s by replying to this "
         "message."
--- a/r2/r2/lib/system_messages.py
+++ b/r2/r2/lib/system_messages.py
@@ -112,6 +112,7 @@ def notify_user_added(rel_type, author, user, target):
 
 def send_mod_removal_message(subreddit, mod, user):
     sr_name = "/r/" + subreddit.name
+    u_name = "/u/" + user.name
     subject = "%(user)s has been removed as a moderator from %(subreddit)s"
     message = (
         "%(user)s: You have been removed as a moderator from %(subreddit)s.  "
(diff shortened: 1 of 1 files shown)


r2/r2/lib/system_messages.py
```diff
@@ -112,9 +112,9 @@ def notify_user_added(rel_type, author, user, target):
 
 def send_mod_removal_message(subreddit, mod, user):
     sr_name = "/r/" + subreddit.name
-    subject = "You've been removed as a moderator from %(subreddit)s"
+    subject = "%(user)s has been removed as a moderator from %(subreddit)s"
     message = (
-        "You have been removed as a moderator from %(subreddit)s.  "
+        "%(user)s: You have been removed as a moderator from %(subreddit)s.  "
```
KeyserSosa: The format replacement on `122-3` also needs to be updated.

KeyserSosa:
💅 This will break as written (see comment) but otherwise the idea of the change works for me.

SmithsonianDSP:
Oh, yeah, I missed adding it to the formatting, there. Thanks, I'm adding it right now. 

Do you have a preference/opinion on whether to use `/u/username` or if just `username` is fine?

KeyserSosa:
I think it would make sense to put the `/u/` in the subject, but leave it out in the message body?

SmithsonianDSP:
Okay, done.

KeyserSosa:
:fish: lgtm

We don't deploy on Fridays as a rule, but I'll get this in on Monday.

SmithsonianDSP:
>We don't deploy on Fridays as a rule

As a developer, myself: I **_completely_** respect that rule. :grinning: 

Monday will be great.

SmithsonianDSP:
But Tuesday also works... 😉

KeyserSosa:
D'oh!  I have failed you...

I'll merge once I get in!

KeyserSosa:
merged 4e7029bdbedd123e8d666fa3b8ac26f6e18ec368

PR #1631, opened 2016-07-01
Adding trial of Honeycomb.io as an analytics provider

maplebed:
We are adding Honeycomb to comment_tree as a trial to begin capturing a sample of reddit data and for reddit to evaluate Honeycomb.

We added lines to example.ini for the API key, dataset name, and sample rate for sending events to Honeycomb. The production config should mirror these changes with real values.

The integration adds a new provider named Analytics and an implementation of that provider for Honeycomb and a no-op implementation named Null. We added an optional argument to Stats.get_timer for the timer creator to elect to send the event to Honeycomb as well as statsd. We added code using analytics to comment_tree.add_comments to show how it should be used.

We didn't set up the entire reddit environment to do a comprehensive test. We'll be in touch to answer any questions.

**Note**: In order for the reddit service to perform a clean shutdown, g.analytics_provider.close() **must** be called from somewhere in the shutdown process - we're not sure where that should be. Without calling close, there will be threads running from the Honeycomb provider that will cause the shutdown process to hang. 

cc: @christineyen @kjoconnor

### DIFF
--- a/r2/example.ini
+++ b/r2/example.ini
@@ -74,6 +74,8 @@ modmail_email_secret = YWJjZGVmZ2hpamtsbW5vcHFyc3R1dnd4eXowMTIzNDU2Nzg5
 mailgun_api_key =
 # request signing
 request_signature_secret = YWJjZGVmZ2hpamtsbW5vcHFyc3R1dnd4eXowMTIzNDU2Nzg5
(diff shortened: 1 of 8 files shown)


bsimpson63:

PR #1633, opened 2016-07-05
POST_site_admin: Remove captcha/ratelimit errors if editing sr settings

13steinj:
There are cases (albeit rare) in which editing settings cause a ratelimit error to occur, even though it does not affect anything except that it gets added to the "saved" status text, such as when editing settings quickly after creation of a subreddit. Now, the error will be preemptively removed in the case that one is editing settings.

There are also cases such as @praw-dev/praw#582 where a captcha error is returned in the json response, so now captcha errors are also to be preemptively removed in the case that one is editing settings.


bsimpson63:

PR #1634, opened 2016-07-07
Controllers: Remove `@vary_pagecache_on_experiments` entries

atiaxi:
On stock Ubuntu 14.04, the automated install script for Ubuntu Linux 14.04 fails with the following error:

`NameError: name 'vary_pagecache_on_experiments' is not defined`

As far as I can tell, this is because that function was removed in 6dbe772e03841041ac2c89d37fda55e02e299789 .  I've removed the lines that still referenced it and the install script now works perfectly (and also the site appears to work)

### DIFF
--- a/r2/r2/controllers/api.py
+++ b/r2/r2/controllers/api.py
@@ -641,7 +641,6 @@ class ApiController(RedditController):
         ))
         return handle_login(**kwargs)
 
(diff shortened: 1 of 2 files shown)


spladug:
Thanks a bunch, Roger!  Merged as 4b2a669d6fe85f85c518a06a905ef1c0cfa57478.

PR #1636, opened 2016-07-07
Fix format of subreddit titles in system messages

13steinj:
Subreddit titles can contain links to other subreddits and or usernames. If these links contain underscores, Snudown parses these incorrectly if they contain underscores because the entire title is wrapped in a `[]()` link block, similar to the issue referenced in 7874382

This resolves the issue mentioned in [this post](https://reddit.com/r/bugs/comments/4rng30/_/).

I've left the traffic message alone since I don't know what title attribute it it references of the top of my head. I'm assuming the submission title; but I'd rather not take the chance.


bsimpson63:

PR #1640, opened 2016-07-10
Set az_selfserve_network_id to 0 to fix installs

13steinj:
Installing reddit off the bat is currently broken because `az_selfserve_netowork_id` is expected to be an integer.

This sets it to 0 so installing reddit would work.


13steinj:
Corrected as of #1644

zeantsoi:
Thanks! Apologies for not catching this PR before opening #1644.

13steinj:
np

PR #1648, opened 2016-07-15
Allow a negative `num` query in /about/sticky

not-an-aardvark:
This allows for links to the *last* sticky submission on a subreddit, using `/r/subreddit/about/sticky?num=-1`. At the moment it is not possible to do this, because `/about/sticky` always returns the first sticky on the subreddit, and `/about/sticky?num=2` will return a 404 error if the subreddit only has one sticky.

Motivation: Some subreddits have a recurring submission that is always stickied (for instance, a daily questions thread). This submission is usually in the first sticky slot when there is no other sticky, but it is occasionally moved to the second sticky slot if another sticky submission takes its place. It would be useful to have a link that always points to the current version of the recurring thread, but this is not possible without being able to link to the last sticky.


bsimpson63:

PR #1649, opened 2016-07-16
LimitUploadSize: map maximum sizes by path and post data

### LINKED ISSUE #1557: Wiki edit requests over 500 KiB rejected for being too large
Despite `/wiki/usernotes` being given [a size limit of 1 MB](https://github.com/reddit/reddit/blob/master/r2/r2/models/wiki.py#L84), submitting a request over 500 KiB/512 KB results in the following 413 error:

I tested by directly editing `/wiki/usernotes` and found [any request over 512 KB is rejected](http://i.imgur.com/VLm2diV.png). No error is shown on the wiki interface, which I assume is due to the error being returned as HTML rather than JSON for errors [raised further down the line](https://github.com/reddit/reddit/blob/master/r2/r2/models/wiki.py#L379).

This error is returned by [the `LimitUploadSize` middleware](https://github.com/reddit/reddit/blob/9c680bdd1d6b5f95e423d509384a5f3c7864d515/r2/r2/config/middleware.py#L368), which I'm assuming it's intercepting and rejecting the request before reaching the wiki model.

I'm not submitting a pull request for now because I'm not sure which approach you would prefer to take and what negative consequences they could have in other parts of the site.

tl;dr: Usernotes are broken and I can't fix them this time. :heart:

13steinj:
Instead of one single maximum upload size, use a mapping to allow for endpoints to have a custom maximum post size.

The default is 500Kib as before, and entries are added to the map to be whitelisted. The path matches against the environ's PATH_INFO, stripped of a leading slash, and all keys selected must match the request's post data. A copy of the environ is sent to the map to keep webob from manipulating it and sending an environ with extra values to the downstream apps.

E: closes #1557

### DIFF
--- a/r2/r2/config/middleware.py
+++ b/r2/r2/config/middleware.py
@@ -21,6 +21,7 @@
 ###############################################################################
 
 """Pylons middleware initialization"""
+from collections import OrderedDict
 import importlib
 import re
 import urllib
@@ -52,6 +53,8 @@ from r2.lib import csrf, filters
 # patch in WebOb support for HTTP 429 "Too Many Requests"
 import webob.exc
 import webob.util
+# patch in LimitUploadSize support for whitelisted variations
+from webob.request import Request
 
 class HTTPTooManyRequests(webob.exc.HTTPClientError):
     code = 429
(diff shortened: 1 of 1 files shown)


bsimpson63:

PR #1651, opened 2016-07-19
Add backdrop-filter and text-smoothing properties to CSS filter + @supports selector

AppleBetas:
When trying to push a change to one of my subreddit's stylesheet, I realized that there were a couple CSS properties missing that may come in handy.

They are:
- `text-smoothing`
- `backdrop-filter`

Now, a bit about `backdrop-filter`. It's friend `filter` was left out of Reddit intentionally due to a [XSS vulnerability in earlier versions of IE](https://github.com/reddit/reddit/pull/1058#issuecomment-76466180). But since `backdrop-filter` is relatively new, it isn't in Internet Explorer, and cannot be exploited. It is safe for use.

Also, with these comes in handy `@supports` (to detect browsers which support specified CSS properties), so I've also added that to the pull request.

Adding these properties would make stylesheets look just a bit better. :)

### DIFF
--- a/r2/r2/lib/cssfilter.py
+++ b/r2/r2/lib/cssfilter.py
@@ -468,7 +468,7 @@ class StylesheetValidator(object):
 
         keyword = strip_vendor_prefix(rule.lower_at_keyword)
 
-        if keyword in ("media", "keyframes"):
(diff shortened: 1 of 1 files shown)


bsimpson63:

PR #1652, opened 2016-07-21
Preserve search syntax between searches.

13steinj:
Search syntax reverts to the default instead of preserving from run to run since it isn't set in the form. This commit adds the syntax to the html form so syntax is preserved between searches. Originally reported [here](http://www.reddit.com/r/bugs/comments/4c3h17/_/)


bsimpson63:

PR #1653, opened 2016-07-22
Update license headers to 2016

StephanMeijer:
Update license headers to 2016


13steinj:
The admins like doing this internally, and on previously related prs they've said as such.

PR #1676, opened 2016-08-19
Fix two bugs regarding ban notes

13steinj:
Currently when making bans permanent the note is not persisted. This persists the note in the "make permanent" button however lets an API client still purge the note if they wish to. Originally reported [here](https://www.reddit.com/r/ModSupport/comments/4yk02f/when_a_ban_is_made_permanent_can_we_keep_the_ban/).

I also found a way in which the ban note can collectively be over 300 characters when using an API client, so I limited it.


bsimpson63:

PR #1682, opened 2016-09-01
Fix bad JS in catch-all JS handler.

indrora:
It's broken.

`r.sendError` isn't defined ever, but it is held in `r.logging.sendError`. This means that (as a side effect) if something IS going wrong, the error logging isn't happening, like this:

image

So I fixed it.


bsimpson63:

PR #1691, opened 2016-09-08
Set meta description for subreddit listings

gg:
The meta description for a subreddit listing is currently set to the [site's tagline](https://github.com/reddit/reddit/blob/master/r2/r2/templates/base.html#L35) (`g.short_description`). Why not set the meta description to the subreddit's `public_description`, so it matches the `og:description`?


bsimpson63:

PR #1697, opened 2016-09-27
Incorrect documentation for GET_rules in subreddit

avinashbot:
It seems that `/r/{subreddit}/rules` isn't a valid endpoint (sorta, it redirects to the main rules). `/r/{subreddit}/about/rules` is the correct path for the subreddit rules (both for frontend and api), which the docstring says is the purpose of the action.

It might also be a good idea to also add the PoliciesController (what the path is incorrectly redirecting to) to the api docs in the future, since they spit out a valid "Policy" object when accessing it via JSON.

### DIFF
--- a/r2/r2/controllers/front.py
+++ b/r2/r2/controllers/front.py
@@ -958,7 +958,7 @@ class FrontController(RedditController):
         return Reddit(content=usertext).render()
 
     @require_oauth2_scope("read")
-    @api_doc(api_section.subreddits, uses_site=True)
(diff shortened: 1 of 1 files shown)


bsimpson63:
🐟 Thanks, looks good.

bsimpson63:
Merged as b900d9b3429bd2aacaeb1b89a76a343e55c920d5

bsimpson63:
Thanks for the fix. Let me know your reddit username if you want an "open sorcerer" award https://www.reddit.com/wiki/awards#open_sorcerer.

PR #1708, opened 2016-10-12
Strip whitespace from usernames on flair lookup and assignment

13steinj:
For the ease of mods copypasting a username from elsewhere, such as an excel spreadsheet, whitespace is left and then the user would not be found. This just strips whitespace from the username before looking up the account name, as done elsewhere such as in `VThrottledLogin`.

### DIFF
--- a/r2/r2/lib/validator/validator.py
+++ b/r2/r2/lib/validator/validator.py
@@ -2736,7 +2736,7 @@ class VFlairAccount(VRequired):
 
     def _lookup(self, name, allow_deleted):
         try:
-            return Account._by_name(name, allow_deleted=allow_deleted)
+            return Account._by_name(name.strip(), allow_deleted=allow_deleted)
         except NotFound:
             return None
 


r2/r2/lib/validator/validator.py
```diff
@@ -2736,7 +2736,7 @@ def __init__(self, item, *a, **kw):
 
     def _lookup(self, name, allow_deleted):
         try:
-            return Account._by_name(name, allow_deleted=allow_deleted)
+            return Account._by_name(name.strip(), allow_deleted=allow_deleted)
```
GHawk1ns: What if `Account._by_name`() handled stripping and any other preconditions check for a valid username instead? That way each caller wouldn't have to worry about this and there'd be some consistency to the preconditions on `name`.

13steinj: > some consistency

I made the change in the effort to keep as much consistency as possible-- in most places it's not stripped and I assume that's a conscious decision. Stripping every time, especially on various get requests where there's a lookup based on previous context (ex cookies), stripping would be pointless. It's only necessary like this on user entry.

bsimpson63:

PR #1712, opened 2016-10-23
Mention karma in CANT_CREATE_SR

allthefoxes:
The most common, and I really mean it, the most common post in /r/help and related subreddits is when users question why they can't make a subreddit even though they have an old account. The error that users get only mentions account age and does not mention any kind of karma requirement. It's in the /r/help FAQ, but we all know no one reads that. 

Just lightly mentioning karma could reduce the traffic on /r/help as well as be more clear.


allthefoxes:
Any update on this? It's a simple string fix. I understand this might break translations, but reddit hasn't seemed to care about its translations for at least a year now.

kjoconnor:
Not sure why you're saying we don't care about translations, we [last updated them on 11/10](https://github.com/reddit/reddit-i18n/commit/1d07738298a8fbb1eaa9e3ce0911dda1499fc8fa)!

Regardless - I'll be merging this in later today. Thanks for the fix!

kjoconnor:
bb1abcccc1bb200eff2c2b6d92f99265054929a1

PR #1719, opened 2016-11-02
[accessibility] Use ARIA regions to group all elements belonging to a thing together

mlang:
reddit.com already uses role="main" to indicate where main content of a page
starts.  However, it is very unconfortable for screen reader users
to navigate from entry to entry. 
This very small patch adds a role="region" attribute around things, which
tells screen readers to indicate to their users that all elements inside that
div actually belong together.  This makes it easy to skip to the next thing.

### DIFF
--- a/r2/r2/templates/printable.html
+++ b/r2/r2/templates/printable.html
@@ -166,7 +166,7 @@ ${self.RenderPrintable()}
    else:
       selflink = ""
  %>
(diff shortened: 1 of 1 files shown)


codeofdusk:
Any idea when this will be merged? Looks like the unit tests are broken...

mlang:
I am going to delete this pull request, as I have done with my reddit
account already.  Several months of no reply are not worth my time.

Thxbye

codeofdusk:
No, please don’t delete it…

 

Bill

PR #1724, opened 2016-11-08
Use relative urls to the same (not the parent) directory in CSS

### LINKED ISSUE #1722: Relative URIs in CSS are malformed.
On local installs, URIS of the nature `../image.ext` such as `../droparrowgray.gif` (a lot of confusion on a heavily modified fork with this file is what led me to find this bug, however I've confirmed this in a non modded fork) is invalid. This is because on reddit.com, Reddit uses a separate domain for static files, `redditstatic.com`. `reddit.<b64hash>.css` is located in the root directory. `../` from this location, because one can not go up once, is simply `redditstatic.com`, and then the image is fetched appropriately.

However, in reality, it's in the same directory. So, `./droparrowgray.gif` would also work. This would work for both local installs and installs that use a separate static domain, as all files are "compiled" into the same directory starting from `r2/public/static`.

`../droparrowgray.gif` does not work on local installs, because going up once from the directory that `reddit.<b64hash>.css` is located would be the `public` build directory, or the root website page (ex `http://reddit.local/`). No such image is located here, only in `http://reddit.local/static/`. A very simple fix would be to replace the `../` with `./` in every applicable location (bar something with the spritesheet iirc).

In an extremely strange twist, not only does it (the `../`) not work (as expected), chrome (and possibly other browsers) get extremely confused (which, may deserve an issue on it's own). For _whatever_ reason, chrome decides that `../droparrowgray.gif` on local installs is equivalent to `http://reddit.local/submit?url=http%3A%2F%2Fdroparrowgray.gif`. I've looked in the JS via `grep` and while I found that general url pattern, (`encodeURIComponent('/?submit=url=http://' + part)` or something along those lines), nothing I've found leads to a reasonable conclusion that something in the JS is doing this. In fact, upon some cursory googling, I haven't found a way to _intentionally_ do this. Or something in the JS is being reused due to the minifier not scoping properly. But still haven't found a way to trick chrome into doing it regardless. So maybe this is a bug on chrome's end.

13steinj:
Relative urls to the parent directory on local installs (build/public) don't have a file to match, so Pylons takes over, and itself fails because the only thing it matches is a potential url (ex `droparrowgray.gif` being a domain name), and then redirects to a submit page to submit it.

Being relative to the same directory is what is needed on local installs, and what is supposed to be done on installs using a static domain. The only reason the real static domain still works is because all files are located in the root directory, and the parent directory of the root directory of a site is considered to be itself.

Two other changes were made to the spritesheet generator and on-the-fly less parser, because the images are located in the parent directory relative to the original `{name}.less/compact.css` file that it gets passed, so it needs to search one directory above now by itself.

For context, the changes to the static files were done via the following script

```
reddit@ubuntu:~/src/reddit/r2$ cd r2/public/static/css && python
>>> from glob import glob
>>> for fpath in (glob('./*.*ss') + glob('./components/*.*ss')):
...     with open(fpath, 'r+') as f:
...         data = f.read().replace('url(../', 'url(./').replace('url("../', 'url("./').replace("url('../", "url('./")
...         f.seek(0); f.truncate(); f.write(data)
```

Resolves #1722

### DIFF
--- a/r2/r2/lib/nymph.py
+++ b/r2/r2/lib/nymph.py
@@ -104,7 +104,7 @@ def _load_spritable_images(css_filename):
                 continue
 
             image_filename, should_stretch, pixel_ratio = _extract_css_info(m)
-            image = Image.open(os.path.join(css_location, image_filename))
+            image = Image.open(os.path.join(css_location, '..', image_filename))
             image_hash = hashlib.md5(image.convert("RGBA").tostring()).hexdigest()
 
             if image_hash not in images:
--- a/r2/r2/public/static/css/compact.css
+++ b/r2/r2/public/static/css/compact.css
@@ -14,8 +14,8 @@ textarea { font-family: inherit; }
 #preload { position: absolute; top: -1000px; left: -1000px; }
 
 /*UI stuff*/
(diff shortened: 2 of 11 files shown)


bsimpson63:

PR #1729, opened 2016-11-14
Typo: Travis doesn't have a "postgres" service, only "postgresql"

13steinj:
Travis is skipping this service as a result.

This doesn't resolve the build errors, which is another issue in and of itself, but I'm semi surprised that the build had worked without this in the first place, unless the corresponding tests were never added.


bsimpson63:

PR #1735, opened 2016-11-25
Search: Fix restrict_sr for domains listings and AllMinus

Pokechu22:
This PR fixes a few issues with the `restrict_sr` property.  Most of the important information is in the commit message:

> For domains listings, it was possible to have results from the wrong domain be included.  Switching to a phrase search fixes that.  See https://redd.it/5ew4ve for more info about, and caveats of, this method.

> For AllMinus, see https://redd.it/5eqiyy for the general context.  This does not need to be explicitly handled for ModMinus, as ModMinus excludes the filtered subreddits from its sr_ids property.  Additionally, ModMinus doesn't need special handling since it inherits the already-handled MultiReddit.

Note that I have _not_ tested this code on a reddit instance (since I don't have cloudsearch), but I did test the individual queries manually (and they work).  I also haven't tested the solr queries at all, but it looks correct per the solr docs.  That entire solr `_get_sr_restriction` seems a bit fishy, though (especially, for instance, `sr == All`), but since I can't test solr I'm not sure if that's actually correct or not.  I'm not sure if solr is actually used anywhere, so I haven't investigated further.

I also reformatted all of cloudsearch.py and solr.py, since they weren't following the style guide.

[Diff excluding whitespace-only changes](https://github.com/reddit/reddit/pull/1735/files?w=1) ([info](https://github.com/blog/967-github-secrets))

### DIFF
--- a/r2/r2/lib/providers/search/cloudsearch.py
+++ b/r2/r2/lib/providers/search/cloudsearch.py
@@ -51,6 +51,7 @@ from r2.lib.providers.search.common import (
 import r2.lib.utils as r2utils
 from r2.models import (
     Account,
+    AllMinus,
     DomainSR,
     FakeSubreddit,
     FriendsSR,
@@ -751,7 +752,7 @@ class LinkSearchQuery(CloudSearchQuery):
     @staticmethod
     def _restrict_sr(sr):
         '''Return a cloudsearch appropriate query string that restricts
-        results to only contain results from self.sr
+        results to only contain results from sr
         
         '''
         if isinstance(sr, MultiReddit):
@@ -760,7 +761,7 @@ class LinkSearchQuery(CloudSearchQuery):
             srs = ["sr_id:%s" % sr_id for sr_id in sr.sr_ids]
             return "(or %s)" % ' '.join(srs)
         elif isinstance(sr, DomainSR):
-            return "site:'%s'" % sr.domain
+            return "site:'\"%s\"'" % sr.domain
         elif isinstance(sr, FriendsSR):
             if not c.user_is_loggedin or not c.user.friends:
                 raise InvalidQuery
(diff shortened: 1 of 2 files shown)


bsimpson63:
Is there any way to force cloudsearch to match the end of a string? That could fix one of the domain edge cases you mention.

Pokechu22:
I'm not aware of any way of doing so for a text field.  The cloudsearch PDF's section on "Searching Text Fields with the Query Parameter" provides 3 subsections on ways to search a field (plus subsections about the `q` and `bq` pareters): boolean operators (doesn't help with this), wildcards (the oposite of what we want), and searching for phrases (what is currently being done).  None of those let you require the text to be the end of the string unfortunately.

However, it does also mention "literal fields", which behave more or less that way (in fact, it's even better): searches against a literal field will only match if it's exact for the entire text (unless a wildcard is used).  If `site` were a literal field, it would behave more or less exactly the way we want.  There's actually several fields (author and subreddit name are the most obvious ones) that could benefit from being literal fields instead of text fields.  But, I don't know if it is possible to retrospectively change the type of the field (or if there are other problems that making it into a literal field would cause) - I understand the searching side of cloudsearch but not the adminy side of it.

bsimpson63:
Cool, thanks for looking into that. I don't think we'll want to change our cloudsearch indexing config or whatever would be required to make `site` a literal field, so this solution seems pretty good.

bsimpson63:
🐟 lgtm, but I probably won't merge the pep8 changes

bsimpson63:
Merged as e5eb480ed38f63eaa4713bbe622bd00755b278bb

bsimpson63:
Thanks for the fix!

Pokechu22:
Thanks for merging!

It's working fairly well on the site, but I just noticed something that I somehow missed - the restrict SR checkbox doesn't appear on /r/all!  I _think_ that when I was testing previously, it might have been visible on `/me/f/all`, but after that feature became available to everyone, it no longer appears.  I might be wrong and it never was there (in which case, derp).

But the functionality still works fine if `restrict_sr=true` is manually added: compare [with](https://www.reddit.com/r/all-wheredidthesodago/search?q=wheredidthesodago&restrict_sr=true) and [without](https://www.reddit.com/r/all-wheredidthesodago/search?q=wheredidthesodago)

Seems like the checkbox was hidden in a9c59254 - can that commit be reverted, since now using `restrict_sr` on /r/all has functionality?

bsimpson63:
Ahh good catch. i don't think we want the checkbox on unfiltered /r/all though.

Pokechu22:
Yea, it would be ideal to not have it on unfiltered /r/all, but I'm not entirely sure if there _is_ an unfiltered /r/all page anymore.  (/r/all always displays "/r/all (filtered)" for me at least, even if the filter is removed).

Probably it would make sense to change the conditions to `not isinstance(c.site, (DefaultSR, AllSR)) or isinstance(c.site, AllMinus)`.  a9c5925 still provides a good reference of where this would need to be changed.

PR #1740, opened 2016-11-29
CSS columns support and related properties

gavin19:
https://developer.mozilla.org/en-US/docs/Web/CSS/columns


DanMan:
Except for IE, they all still need their prefixes, as the MDN page shows. Not sure if this matters.

gavin19:
The parser allows for [prefixing](https://github.com/gavin19/reddit/blob/62f7eb12accdef2a1e66cddbbd75e99ca0612af3/r2/r2/lib/cssfilter.py#L62) properties. I tested the changes locally before making the request.

bsimpson63:

PR #1742, opened 2016-12-03
Ensure PyCrypto v2.6.1 is used.

EdOverflow:
Stable versions of PyCrypto prior to 2.6.1 are vulnerable to [CVE-2013-1445](https://lists.dlitz.net/pipermail/pycrypto/2013q4/000702.html).


JordanMilne:
We don't typically enforce the usage of specific versions in our `setup.py` because we manage all of our python dependencies via our `apt` repo. You're not intended to have `setuptools` install the dependencies from pypi for you, and I'd be surprised if it worked correctly. 

Just going to close this since we don't maintain this repo anymore.

PR #1749, opened 2016-12-27
Implement Content-Security-Policy (CSP)

EdOverflow:
Content Security Policy (CSP) is a security mechanism that allows websites to control how browsers process their pages. In essence, sites can restrict what types of resources are loaded and from where. CSP policies
can be used to defend against cross-site scripting, prevent mixed content issues, as well as report violations for investigation.

I am fairly sure that I have not supplied enough policies, therefore if you encounter warnings of any sort please report back and I will add them to the list.

For more on CSP please refer to the following page: https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP


bsimpson63:

PR #1755, opened 2017-01-19
Fix severe jank on loading additional comments

larsjohnsen:
`.hide().show()` serves no apparent purpose, but causes a reflow since `hide` forces a style recalculation.

The reason why this is so severe, is that `add comment to DOM` then `check whether hidden` is repeated for each comment. Every time the DOM is changed, `jQuery.hide` forces the browser to do style recalculations on the parent `.siteTable` in order to check hidden status. This is especially expensive at the bottom "load more comments" since that recalculation will involve every comment thread.

Before:
image

After:
image


bsimpson63:

Issue #1758, opened 2017-01-24
Bug in contest mode prevents later users' posts from being shown


kboone:
If a contest mode thread goes past the maximum number of comments for a page (eg: 200), it will ignore any comments past the 200th one posted when randomizing the order. For example, see https://www.reddit.com/r/churning/comments/5ll4xi/official_discover_it_referral_thread/. Any comment that wasn't posted within the first day on that thread does not show up in the randomized comments unless you hit load more.

This could potentially be abused... a malicious user could post a lot of early replies to a thread, making it so that other users never have their posts shown in contest mode. For example, if someone was doing a giveaway and using contest mode to choose a winner, a malicious user could guarantee a win by making many rapid posts regardless of how many future posts there are. I sent this to security@reddit.com originally but was told that it was OK to post here.

I dug into the code to try to figure out what was going on. It looks like the issue is in CommentOrdererBase.get_comment_order in builder.py. That function returns a list of comments sorted by "_confidence" which I think returns the same number for everything in contest mode. It then picks the top 200 comments from that which defaults to the comments by date. The shuffling is then applied in CommentBuilder._make_wrapped_tree which only is run on the top 200 comments.

One solution would be to sort by some random number rather than by _confidence in comment mode which would choose the 200 comments randomly rather than by date. I don't know how that would interact with your caching system. The random seed would probably need to be kept around to not mess up the shuffle order across pages.

Enuratique:
To make @kboone's post more clear. If a thread's total comments (including removed comments) exceeds the maximum limit, then contest mode isn't used at all, and ALL posts are sorted by date ascending. Quite the opposite intended effect of contest mode. Considering this manifests itself after just 200 comments, this issue probably affects thousands of threads on a daily basis at Reddit.

tomkel:
@bsimpson63 can you take a look at this?

13steinj:
There's more to it than that (at least last i checked which was months ago)-- technically contest mode's "randomization" can be reversed via (a lot of) hits to the page, sequentially increasing the "limit" parameter, because the randomization just takes the "best", generates the tree up to the limit, then randomizes (or gets the "limit" amount of comments and then randomizes, then generates the tree, whichever, the point stands). I messaged discreetly about this and was told it would be looked in to. Suffice to say, it wasn't at minimum 3 months later, and a cursory look at the commit history says no change was made either by now.

[...]

PR #1762, opened 2017-02-09
Fix contest mode.

tomkel:
This assigns all comments a random score in contest mode, before they get split into `comment_tuples` and `missing_root_comments`. The current method doesn't include the `missing_root_comments` in the shuffle.

See #1758.

### DIFF
--- a/r2/r2/models/builder.py
+++ b/r2/r2/models/builder.py
@@ -943,7 +943,7 @@ SORT_OPERATOR_BY_NAME = {
     "qa": operators.desc('_qa'),
     "hot": operators.desc('_hot'),
     "top": operators.desc('_score'),
-    "random": operators.shuffled('_confidence'),
+    "random": operators.shuffled('_random'),
 }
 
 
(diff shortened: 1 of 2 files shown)


r2/r2/models/builder.py
```diff
@@ -943,7 +943,7 @@ def update_candidates(self, candidates, sorter, to_add=None):
     "qa": operators.desc('_qa'),
     "hot": operators.desc('_hot'),
     "top": operators.desc('_score'),
-    "random": operators.shuffled('_confidence'),
+    "random": operators.shuffled('_random'),
```
bsimpson63: I think this dict is only used to figure out which precomputed comment order to use. The one you want to change is in `SortMenu` https://github.com/reddit/reddit/blob/master/r2/r2/lib/menus.py#L560

bsimpson63: Are you able to test this change?

It looks like the rest of these strings ("_date", "_controversy", etc) have corresponding properties like https://github.com/reddit/reddit/blob/master/r2/r2/lib/db/thing.py#L685. When are these properties accessed, and do you need to add one for `_random`?

PR #1764, opened 2017-02-15
Improve Atom feeds

Cj-Malone:
Add extra metadata to the Atom entries, link to the comments (Atom + HTML) and related links when not a self post or in a subreddit like /r/all.

Hopefully you could drop https://github.com/reddit/reddit/blob/6484a52fad73ea747ab21edf4182b67929e15b6f/r2/r2/templates/link.xml#L68 line 68-92 when client support is a bit better.

Note I've not tested this


bsimpson63:

PR #1766, opened 2017-02-22
CSS object-fit support

gavin19:
https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit


bsimpson63:

PR #1769, opened 2017-03-10
Added transparency to the 404 images

robotortoise:
Just what it sounds like. I added transparency to the 404 images. Don't
know if they're still used, but....meh. Can't hurt, right?


Pokechu22:
Not sure if all of those images are used, but some definitely are.  Do note that there are a (few) more images on [reddit/error-pages](https://github.com/reddit/error-pages) which are also occasionally used (but hopefully shouldn't ever be).  Otherwise, looks good to me (but note that I'm not a Reddit employee).

bsimpson63:

PR #1544, opened 2016-03-05
Fix issue when 'compiling' CSS

huyphan:
The change is meant to fix the error `No rule to make target 'build/public/static/reddit.css'` when running `make` or `make css`:

```
$ make clean && make css
[+] including definitions from Makefile.py
rm -f ./r2/lib/mr_tools/_mr_tools.c ./r2/lib/wrapped.c ./r2/lib/utils/comment_tree_utils.c ./r2/lib/utils/_utils.c ./r2/lib/sgm.c ./r2/lib/db/_sorts.c ./r2/lib/mr_tools/_mr_tools.so ./r2/lib/wrapped.so ./r2/lib/utils/comment_tree_utils.so ./r2/lib/utils/_utils.so ./
r2/lib/sgm.so ./r2/lib/db/_sorts.so
rm -f r2/lib/generated_strings.py
rm -rf build/static-buildstamp build/public/static  build/mangle-buildstamp
[+] including definitions from Makefile.py
cp -ruTL r2/public build/public
touch build/static-buildstamp
make: *** No rule to make target 'build/public/static/reddit.css', needed by 'css'.  Stop.
```

PR #1775, opened 2017-03-14
Add new user agent detector for Android

heyhuyen:
Adds a new detector for the current user agent string that the Android app is using. Renames the old one so we can keep capturing old user agents.

Format (from Android strings.xml): `Reddit/Version %1$s/Build %2$s/Android %3$s`
Example: Reddit/Version 2.1.11/Build 20111/Android 6.0.1

### DIFF
--- a/r2/r2/lib/utils/reddit_agent_parser.py
+++ b/r2/r2/lib/utils/reddit_agent_parser.py
@@ -103,7 +103,7 @@ class RedditIsFunDetector(RedditBrowser):
 
 
 @register_detector
-class RedditAndroidDetectorOld(RedditBrowser):
+class OldRedditAndroidDetector(RedditBrowser):
     is_app = True
     look_for = 'RedditAndroid'
     name = 'Reddit: The Official App'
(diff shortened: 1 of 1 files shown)


heyhuyen:
👓 : @KTAtkinson (Can you review this? I'm not sure who should review this, so I just tagged you since you touched this file most recently)

r2/r2/lib/utils/reddit_agent_parser.py
```diff
@@ -103,14 +103,24 @@ class RedditIsFunDetector(RedditBrowser):
 
 
 @register_detector
-class RedditAndroidDetector(RedditBrowser):
+class RedditAndroidDetectorOld(RedditBrowser):
```
KTAtkinson: I would rename this `OldRedditAndroidDetector`, then all the detectors end with a common post-fix.

heyhuyen: fixed.

heyhuyen:
💇: @KTAtkinson 
Also, should I be concerned about the CI build failures?

KTAtkinson:
:fish:

It looks like that error is a css compilation error, so you shouldn't need to worry about it. There is a test for this code, run it and ensure that it passes.

Pokechu22:
>  Also, should I be concerned about the CI build failures?

CI's been broken for a while unfortunately; probably, it's not your fault.  See PR #1544 (may also want to look at #1729 and some of the other PRs)

PR #1796, opened 2017-04-24
Raise `NotImplementedError` instead of `NotImplemented`.

chrisgavin:


r2/r2/lib/mr_tools/mr_tools.py
```diff
@@ -88,7 +88,7 @@ def __init__(self):
         pass
 
     def process(self, values):
-        raise NotImplemented
```
elnuno: I think this is not the 'right' (as in 'intended by the language') thing to raise. It seems that `NotImplemented` is meant to be returned from dunder methods, while [NotImplementedError](https://docs.python.org/3/library/constants.html#NotImplemented) is the one that should be raised.


Of course, it makes no sense to 'fix' this if the code expects catching `NotImplemented`.

chrisgavin: Ah yes, you are completely right. Given that the code currently throws a type error (because `NotImplementedType` is not derived from `BaseException`) I don't think anyone is depending on the current behavior.

bsimpson63:

PR #1797, opened 2017-04-24
Remove unneeded else in `get_step_state`.

chrisgavin:
This `else` seems like it could be confusing. There is no `break` in the loop, so the `else` will always be executed if the `return` in the loop was not triggered.


bsimpson63:

PR #1843, opened 2017-08-14
Rework AdminAwardGive to use description default

tueksta:
* No double initializing of the empty descriptor anymore.
* Time only retrieved if actually needed
* Better readability


bsimpson63:

Issue #1789, opened 2017-04-07
oEmbed endpoint does not support CORS or JSONP


cperryk:
Although the full Reddit API supports CORS and JSONP, the [oEmbed endpoint](https://github.com/reddit/reddit/wiki/oEmbed) does not appear to support either, making it impossible for a client-side app to use it without some kind of proxy. I've whipped up a quick JS fiddle to demo this issue [here](https://jsfiddle.net/cperryk/Lm4gozh0/1/).

PR #1848, opened 2017-08-28
add jsonp support to oembed endpoint

nelsonpecora:
This is a naive solution to #1789, but I can't really figure out how JSONP support is working for other endpoints (`set_content_type` from `reddit_base.py` doesn't seem to be called anywhere, and nothing else is explicitly setting `application/javascript` content type for JSONP requests).

I'm not super great at python (though I've touched the oembed endpoint before 😅 ), so I'd appreciate pointers / more elegant solutions to this, thanks!


bsimpson63:

Issue #1850, opened 2017-09-02
20 undefined names


cclauss:
https://travis-ci.org/reddit/reddit/jobs/271121841#L477

PR #1852, opened 2017-09-02
scripts/hashdist.py import errno

cclauss:
Fixes one issue in #1850  `errno.EPIPE`is used on line 86.

https://docs.python.org/2/library/errno.html


cclauss:
Closing because of https://www.reddit.com/r/changelog/comments/6xfyfg/an_update_on_the_state_of_the_redditreddit_and/

PR #1853, opened 2017-09-02
Define EXPIRES_DATE_FORMAT = "%Y-%m-%dT%H:%M:%S"

cclauss:
Fixes one issue in #1850   `EXPIRES_DATE_FORMAT` is used on line 83 but is not defined.  The definition was copied from https://github.com/reddit/reddit/blob/master/r2/r2/controllers/promotecontroller.py#L159


cclauss:
Closing because of https://www.reddit.com/r/changelog/comments/6xfyfg/an_update_on_the_state_of_the_redditreddit_and/

